PR #18, opened 2023-11-11
feat: add list type

skyzh:
apart from group id placeholder, user's plan node type should also have a slot for list type.


skyzh:
this pull request also replaces `OneOrMany` to the list type in rule matcher.

PR #19, opened 2023-11-11
feat: add rule matcher DSL

### LINKED ISSUE #6: rule DSL
we already have the intermediate representation of rule matchers, and therefore it would be easy to implement a Rust macro that converts a rule DSL into the IR. one proposal:

```
(join inner !a !b !cond) -- defines a matcher on join a and b, for join commute rule
(join inner (join inner !a !b !condab) !c !condabc -- defines a matcher on two joins, for join assoc rule
```

`!a`, `!b`, `!cond` will be picked up and the rule can use them to compose the new plan node.

skyzh:
close https://github.com/cmu-db/optd/issues/6

this pull request adds a Rust-marco based rule-DSL for optd. see `rules::joins` for more example. need unit testing so as to ensure the pick order and retrieve order is the same.


skyzh:
need unit test, not sure if Rust will reorder field assignments...

PR #24, opened 2023-11-22
feat: add physical plan conversion to datafusion

skyzh:


skyzh:
not sure if the converted physical plan really runs -- need testing 🤪

PR #42, opened 2023-12-27
fix: arrow_schema type mismatched

xiaguan:
When I tried to run following command in `README.md`,
```
cargo run --release --bin optd-adaptive-tpch-q8
```
Env:
Ubuntu 22.04 
rust : stable

The compiler error is
```
mismatched types
`arrow_schema::Schema` and `Schema` have similar names, but are actually distinct types
perhaps two different versions of crate `arrow_schema` are being used?
```

After I replaced `arrow_schema` with `datafusion::arrow::datatypes`.  I could run the example.


skyzh:
hmm... this may be caused by inconsistent version of arrow version that I used locally and you are using. I guess they are not strictly following the semver requirement. I'll look into this later and pin everything to the correct version.

skyzh:
just added `Cargo.lock` to the repo -- this is a bad practice but I believe optd repo will be the only user of optd-core for now so it does not matter.

skyzh (approved):

PR #44, opened 2024-01-31
[Fix] apply rule works for expr instead of group

AveryQi115:
# Main Idea

ApplyRuleTask should work for expr if mark_rule_is_fired only works for expr_id for removing duplicated applied rules.


skyzh (approved):
MExpr in Cascades = RelMemoNode in optd, so this looks correct

PR #49, opened 2024-02-03
feat: [Logical Optimizer]Eliminate Join Rule

AveryQi115:
# Major Changes
Add an eliminate join rule which 
1. converts inner join to cross join when filter always equals to true
2. converts inner join to empty relation (remove the join) when filter always equals to false

## Rule Type
Heuristics(always apply), Transformation Rule(logical to logical)

# TODO
- [x] Create Empty Relation Support
- [x] Integrate with cross join type without condition

### DIFF
--- a/optd-datafusion-bridge/src/from_optd.rs
+++ b/optd-datafusion-bridge/src/from_optd.rs
@@ -345,7 +345,7 @@ impl OptdPlanContext<'_> {
         }
 
         let join_type = match node.join_type() {
-            JoinType::Inner | JoinType::Cross => datafusion::logical_expr::JoinType::Inner,
+            JoinType::Inner => datafusion::logical_expr::JoinType::Inner,
             JoinType::LeftOuter => datafusion::logical_expr::JoinType::Left,
             _ => unimplemented!(),
         };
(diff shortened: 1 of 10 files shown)


AveryQi115:
@yliang412 yuchen, could you check whether the pr has any conflicting logic with your cross join support and arguments for logical optimizer? If no conflict, I'll merge it tonight.

yliang412:
Doing some check right now

optd-datafusion-bridge/src/from_optd.rs
```diff
@@ -345,7 +345,7 @@ impl OptdPlanContext<'_> {
         }
 
         let join_type = match node.join_type() {
-            JoinType::Inner => datafusion::logical_expr::JoinType::Inner,
+            JoinType::Inner | JoinType::Cross => datafusion::logical_expr::JoinType::Inner,
```
yliang412: `CrossJoin` already handled on line 343?

yliang412 (approved):
Overall LGTM

optd-datafusion-bridge/src/from_optd.rs
AveryQi115: Removed the Cross in this branch. Thank you for catching it!

PR #50, opened 2024-02-03
test: add cross-product test and enable more explain options

yliang412:
**Changes**
- [x] enable more explain subtask options  
- [x] adds a cross-product test
- [x] document `optd-sqlplannertest` usage
- [x] add flags/tasks to use datafusion logical optimizer.

The `join_enumerate` test now would keep using the logical optimizer until we have full support from optd's side.

### DIFF
--- a/optd-sqlplannertest/src/lib.rs
+++ b/optd-sqlplannertest/src/lib.rs
@@ -102,7 +102,38 @@ impl sqlplannertest::PlannerTestRunner for DatafusionDb {
                 let result = self.execute(&format!("explain {}", test_case.sql)).await?;
                 for subtask in task["explain:".len()..].split(",") {
                     let subtask = subtask.trim();
-                    if subtask == "join_orders" {
+                    if subtask == "logical_datafusion" {
+                        writeln!(
+                            r,
+                            "{}",
+                            result
+                                .iter()
+                                .find(|x| x[0] == "logical_plan after datafusion")
+                                .map(|x| &x[1])
+                                .unwrap()
+                        )?;
+                    } else if subtask == "logical_optd" {
+                        // println!("logical!!!");
+                        writeln!(
+                            r,
+                            "{}",
+                            result
+                                .iter()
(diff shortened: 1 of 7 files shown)


skyzh:
looks good -- btw, it might be better to test the rules with the heuristics optimizer as its output is stable and not cost-based (not sure if it still works now...)

yliang412:
Test output

```
   Compiling optd-sqlplannertest v0.1.0 (/Users/liangyc/workplace/cmu-db/optd/optd-sqlplannertest)
    Finished test [unoptimized + debuginfo] target(s) in 1.27s
    Starting 2 tests across 3 binaries
        PASS [   0.041s] optd-sqlplannertest::planner_test cross_product
        PASS [   3.874s] optd-sqlplannertest::planner_test join_enumerate
```

skyzh (approved):
thanks for adding the docs!

optd-sqlplannertest/src/lib.rs
```diff
@@ -79,53 +104,125 @@ impl DatafusionDb {
...
+    async fn task_execute(&mut self, r: &mut String, sql: &str, with_logical: bool) -> Result<()> {
+        use std::fmt::Write;
+        let result = self.execute(&sql, with_logical).await?;
+        writeln!(r, "{}", result.into_iter().map(|x| x.join(" ")).join("\n"))?;
+        writeln!(r)?;
+        Ok(())
+    }
+
+    /// Executes the `explain` task.
+    async fn task_explain(
```
Sweetsuro, line 118: Is there a cleaner/more modular way to write this function? I wonder if we'll want to add more subtasks later for stuff other than join orders

Sweetsuro (approved):
Overall LGTM

optd-sqlplannertest/src/lib.rs
yliang412: I think right now everything is working around with yaml -> string parsing. We could refactor this if we find the need to add more subtasks for explain.

PR #52, opened 2024-02-08
feat: [LogicalOptimizer] add support for empty relation

AveryQi115:
# Major Changes
- Add support for empty relation
- Schema conversion, from optd schema properties to datafusion::arrow::Schema
- cost calculation for empty relation
- datafusion_bridge support for empty relation conversion between datafusion and optd
- sql planner test for empty relation

# EmptyRelation Usecases
```
select 64+1;
```
The above query creates a datafusion plan Projection -> EmptyRelation(produce_one_row = true)
```
select 64+1 from t1 inner join t2 on false;
```
The above query creates a datafusion plan EmptyRelation with appropriate transformation rules.

# Alternative Design
EmptyRelation can be designed as a kind of logical properties for TableScan. But this design makes executing queries like `select 64+1;` hard.

# Future Task
## schema
Current schema support is incomplete. 
- Not all group in optd cascades optimizer have schema, which means we cannot get schema from a general PlanNode (PhysicalCollector cannot get schema from the PlanNode).
- Optd Schema only contains type information, lacks name and is_nullable, which is incompatible with arrow::Schema
- There's no direct schema conversion from optd schema to arrow::schema, from_optd function uses datafusion table scan's schema as the schema for scan node and pull up the schema to its parent node. However, empty relation needs schema conversion from optd schema to arrow::schema when outputting the physical plan as there are no table name exist in datafusion for empty relation.

optd schema should supports name, is_nullable and more types to make a 1-1 mapping to arrow::Schema.

## JoinElimination Rule
https://github.com/cmu-db/optd/pull/49

### DIFF
--- a/optd-datafusion-repr/Cargo.toml
+++ b/optd-datafusion-repr/Cargo.toml
@@ -7,8 +7,6 @@ edition = "2021"
 
 [dependencies]
 anyhow = "1"
-arrow-schema = "*"
-datafusion = "32.0.0"
 num-traits = "0.2"
 num-derive = "0.2"
 tracing = "0.1"
--- a/optd-datafusion-repr/Cargo.toml
+++ b/optd-datafusion-repr/Cargo.toml
@@ -7,6 +7,8 @@ edition = "2021"
 
 [dependencies]
 anyhow = "1"
+arrow-schema = "*"
+datafusion = "32.0.0"
 num-traits = "0.2"
 num-derive = "0.2"
 tracing = "0.1"
--- a/optd-datafusion-repr/src/properties/schema.rs
+++ b/optd-datafusion-repr/src/properties/schema.rs
@@ -6,13 +6,38 @@ use std::{
 use optd_core::property::PropertyBuilder;
 
 use crate::plan_nodes::{ConstantType, OptRelNodeTyp};
+use datafusion::arrow::datatypes::{DataType, Field, Schema as DatafusionSchema};
 
 #[derive(Clone, Debug)]
 pub struct Schema(pub Vec<ConstantType>);
 
+// TODO: add names, nullable to schema
 impl Schema {
     pub fn len(&self) -> usize {
         self.0.len()
+    }
+}
+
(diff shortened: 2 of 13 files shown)


optd-datafusion-repr/Cargo.toml
```diff
@@ -7,6 +7,8 @@ edition = "2021"
 
 [dependencies]
 anyhow = "1"
+arrow-schema = "*"
```
jurplel: Any reason we're not pinning `arrow-schema` to a specific version?

optd-datafusion-repr/src/properties/schema.rs
```diff
@@ -6,16 +6,41 @@ use std::{
...
 pub struct Schema(pub Vec<ConstantType>);
 
+// TODO: add names, nullable to schema
 impl Schema {
     pub fn len(&self) -> usize {
         self.0.len()
     }
 }
 
+impl Into<DatafusionSchema> for Schema {
```
jurplel: Is there a way to remove this implementation from this repr crate, and add it to the bridge? You should be able to wrap the struct in another type and implement the trait on that. 

I feel that this might be cleaner than introducing a datafusion dependency to the repr crate (since I think we want to go the direction of decoupling this repr from datafusion eventually)

optd-datafusion-repr/src/properties/schema.rs
```diff
@@ -6,16 +6,41 @@ use std::{
 use optd_core::property::PropertyBuilder;
 
 use crate::plan_nodes::{ConstantType, OptRelNodeTyp};
+use datafusion::arrow::datatypes::{DataType, Field, Schema as DatafusionSchema};
 
 #[derive(Clone, Debug)]
 pub struct Schema(pub Vec<ConstantType>);
 
+// TODO: add names, nullable to schema
```
jurplel, line 8: Can we make an issue for this?

AveryQi115: Will create one and raise a pr soon.

PR #53, opened 2024-02-09
feat: integrate statistics into cost model

Gun9niR:


optd-core/Cargo.toml
```diff
@@ -14,3 +14,10 @@ ordered-float = "4"
 tracing-subscriber = "0.3"
 pretty-xmlish = "0.1"
 itertools = "0.11"
+parquet = "50.0.0"
```
skyzh, line 17: optd core should be independent of repr, therefore, I don’t think importing these crates are necessary.

optd-core/src/gungnir/stats.rs
```diff
@@ -0,0 +1,94 @@
...
+use std::fs::File;
+
+use super::tdigest::TDigest;
+
+#[derive(Default)]
+pub struct Stats {
+    pub table_stats: HashMap<String, TableStats>,
+}
+
+pub struct TableStats {
```
skyzh, line 23: Consider finding a generic way of representing stats and let repr implement it

skyzh:
License-wise, I would prefer keeping things consistent with the repo license (CMU DB Group 2023-2024 under MIT). It should be easy to use git blame or GitHub contribution tools to attribute contributions afterwards.

skyzh:
btw, is statistics a type of logical property? (Does Calcite consider it a property or a separate thing?)

PR #54, opened 2024-02-10
feat: Eliminate Filter Rule

jurplel:
TODO: Write documentation & write tests

Depends on #52


yliang412:
I think this is eliminate filter rule right?

jurplel:
Yes, thanks for the correction

optd-core/src/cascades/memo.rs
```diff
@@ -198,6 +222,12 @@ impl<T: RelNodeTyp> Memo<T> {
         props
     }
 
+    pub fn clear_exprs_in_group(&self, group_id: ReducedGroupId) {
+        let group = self.groups.get(&group_id).expect("group not found");
```
skyzh: this does not clear anything 🤣 either directly remove the group id from the hash map, or get a mutable reference to the entry

jurplel: Whoops, that was extremely silly of me. Fixed!

yliang412 (approved):
LGTM! Thanks for the code comments!

PR #55, opened 2024-02-11
feat: linting + testing GitHub Action

patrickhyw:
- Added a GitHub Action (`tests.yaml`) for linting and testing
- Fixed all `clippy` errors
- Commented broken CLI tests
- Added CLI test to run `tpch/test.sql`


.github/workflows/tests.yaml
```diff
@@ -0,0 +1,13 @@
+name: tests
```
jurplel: Maybe a more descriptive name, like "Clippy and Tests" (since this workflow runs both)

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,13 @@
...
+
+on:
+  pull_request
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - run: cargo clippy --all-targets --all-features -- -D warnings
```
jurplel: It would be nice to have names here for the Github GUI

e.g.
```
      - name: Run cargo clippy
         run: cargo clippy --all-targets --all-features -- -D warnings
```

jurplel (approved):

PR #57, opened 2024-02-11
feat: support more integer types

Gun9niR:
- Support more DataFusion integer types.
- Instead of converting every integer type to `Int64`, define more variants of integers.
- Add SQL planner test for integer constant predicate.

This fix only works when DF's optimizer rules are not cleared. DF treats integer constants as Int64, therefore to compare columns with other integer types with the constant it needs to be cast to Int64, which is currently not supported by optd.

To implement `Cast` expression, we will need to support multiple values in `RelNode`'s `value` field.

### DIFF
--- a/optd-core/src/rel_node.rs
+++ b/optd-core/src/rel_node.rs
@@ -27,7 +27,14 @@ pub trait RelNodeTyp:
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub enum Value {
-    Int(i64),
+    UInt8(u8),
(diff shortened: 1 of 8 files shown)


AveryQi115 (approved):
LGTM! The schema pr will also based on these added types

jurplel:
lgtm!

jurplel (approved):

yliang412 (approved):
LGTM. Have you tested these without logical optimizer enabled? Did those break anything?

PR #58, opened 2024-02-11
feat: Support limit operations

jurplel:


optd-datafusion-repr/src/cost/base_cost.rs
```diff
@@ -107,6 +107,11 @@ impl CostModel<OptRelNodeTyp> for OptCostModel {
                 Self::cost(row_cnt, 0.0, row_cnt)
             }
             OptRelNodeTyp::PhysicalEmptyRelation => Self::cost(0.5, 0.01, 0.0),
+            OptRelNodeTyp::PhysicalLimit => {
```
yliang412, line 110: Just wondering how is this determined?

yliang412 (approved):
LGTM, left a small comment trying to understand the cost for limit

optd-datafusion-repr/src/cost/base_cost.rs
jurplel: I don't fully understand the cost system in play here just yet, but there are a few things to notice:
-The `row_cnt` is from the child tuple
- I multiply by selectivity as a very rough *estimate* for what the cost might be 
    - usually LIMIT clauses are quite selective, so that is my sketchy reasoning for 0.001. Also, it is the same selectivity as FILTER.
- the computation cost for this one I just set to `row_cnt`. No reasoning really. Looking again at other examples, I think that should probably be compute_cost from the child, so I will change it.

There is probably a cleverer way to do this by actually reading LIMIT, but I think it might be best to leave that for a real cost model.

PR #60, opened 2024-02-11
feat: Eliminate Limit Rule

jurplel:
Depends on #58


jurplel:
I am not sure how to construct a query that will test this case:
> Limit with skip 0 and no fetch -> Eliminate from the tree

any ideas?

AveryQi115 (approved):
Overall LGTM. I think for testing skip=0 and fetch is none datafusion repo creates fake logical plan with limit as the children node of the limit. If we want to do the same testing using sql, we might need to add support for subqueries or unions or other complicated nested queries to get this internal states...

AveryQi115:
> I am not sure how to construct a query that will test this case:
> 
> > Limit with skip 0 and no fetch -> Eliminate from the tree
> 
> any ideas?

Not sure if current optd can support things like `select * from (select * from t1 limit 5) limit 6;`

jurplel:
> `select * from (select * from t1 limit 5) limit 6;`
This query does work, but it doesn't test the case that I was referring to. I don't know of any way to make a SQL query that will make skip: 0 and fetch: None.

```
| physical_plan after optd                         | PhysicalLimit { skip: 0, fetch: 6 }                                     |
|                                                  | └── PhysicalProjection { exprs: [ #0, #1 ] }                            |
|                                                  |     └── PhysicalLimit { skip: 0, fetch: 5 }                             |
|                                                  |         └── PhysicalProjection { exprs: [ #0, #1 ] }                    |
|                                                  |             └── PhysicalScan { table: t2 }                              |
```

I'll go ahead and merge this for now but let me know if we have a nice way to test with fabricated query plans.

PR #62, opened 2024-02-12
feat: cargo fmt and actions to force format

AveryQi115:
- cargo fmt current code
- actions CI include: format check, clippy check, cargo test


skyzh:
i would not recommend auto-fix fmt. just use `cargo fmt --check` would be good enough :)

skyzh:
like in this case, it seems that you cannot retrieve the token...?

AveryQi115:
> like in this case, it seems that you cannot retrieve the token...?

I think it is related to nightly dependencies issue?

skyzh:
i think the action you are using in rustfmt/format needs a GitHub token to add commits to the pull request.

actually i have an action file that are used across almost all my Rust projects: https://github.com/skyzh/type-exercise-in-rust/blob/main/.github/workflows/ci.yml

skyzh:
and you can pin https://github.com/cmu-db/optd/blob/main/rust-toolchain to a specific version (1.75 is the latest one?) so that there're no more fmt issues

PR #64, opened 2024-02-13
test: add tpc-h q8 and fix expression parsing

skyzh:
TPC-H Q8 has been broken since cross join support b/c it adds join filter support but some expressions were not handled. This pull request adds tests for TPC-H Q8 and implements between + cast expressions. Note that the type inference is broken and someone should add a `Type` rel node to df-repr in order to process types correctly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -154,7 +154,7 @@ dependencies = [
  "arrow-json",
  "arrow-ord",
  "arrow-row",
- "arrow-schema 47.0.0",
+ "arrow-schema",
  "arrow-select",
  "arrow-string",
 ]
@@ -168,7 +168,7 @@ dependencies = [
  "arrow-array",
  "arrow-buffer",
  "arrow-data",
- "arrow-schema 47.0.0",
+ "arrow-schema",
  "chrono",
  "half",
  "num",
@@ -183,7 +183,7 @@ dependencies = [
  "ahash",
  "arrow-buffer",
  "arrow-data",
- "arrow-schema 47.0.0",
+ "arrow-schema",
  "chrono",
  "chrono-tz",
  "half",
@@ -211,7 +211,7 @@ dependencies = [
  "arrow-array",
  "arrow-buffer",
  "arrow-data",
- "arrow-schema 47.0.0",
+ "arrow-schema",
  "arrow-select",
  "chrono",
  "comfy-table",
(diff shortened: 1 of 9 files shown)


AveryQi115:
^ Note that the type inference is broken and someone should add a Type rel node to df-repr in order to process types correctly.
What does this mean?

skyzh:
> ^ Note that the type inference is broken and someone should add a Type rel node to df-repr in order to process types correctly. What does this mean?

1. Cast needs to be represented as cast(child_expr, target_type). Given that we don't have a `Type` RelNode, we use `Value(default_value)` to represent a type.
2. Decimal has precisions and we did not store them. We hardcode it to be (15, 2) when converting optd representation out to DF representation.

```
               let expr = CastExpr::from_rel_node(expr.into_rel_node()).unwrap();
                 let child = Self::conv_from_optd_expr(expr.child(), context)?;
                 let data_type = match expr.cast_to() {
                     Value::Bool(_) => DataType::Boolean,
                     Value::Decimal128(_) => DataType::Decimal128(15, 2), /* TODO: AVOID HARD CODE PRECISION */
                     Value::Date32(_) => DataType::Date32,
                     other => unimplemented!("{}", other),
                 };
```

AveryQi115:
> > ^ Note that the type inference is broken and someone should add a Type rel node to df-repr in order to process types correctly. What does this mean?
> 
> 1. Cast needs to be represented as cast(child_expr, target_type). Given that we don't have a `Type` RelNode, we use `Value(default_value)` to represent a type.
> 2. Decimal has precisions and we did not store them. We hardcode it to be (15, 2) when converting optd representation out to DF representation.
> 
> ```
>                let expr = CastExpr::from_rel_node(expr.into_rel_node()).unwrap();
>                  let child = Self::conv_from_optd_expr(expr.child(), context)?;
>                  let data_type = match expr.cast_to() {
>                      Value::Bool(_) => DataType::Boolean,
>                      Value::Decimal128(_) => DataType::Decimal128(15, 2), /* TODO: AVOID HARD CODE PRECISION */
>                      Value::Date32(_) => DataType::Date32,
>                      other => unimplemented!("{}", other),
>                  };
> ```

Sure, we'll add the support soon.

AveryQi115 (approved):
LGTM

PR #66, opened 2024-02-13
feat: implment column ref as logical property

Gun9niR:
Derive base table column ref for logical nodes, so that we can know which columns the `ColumnRef` objects in a predicate refers to.

Tested manually on the two demos. Not sure how to unit test it.


skyzh (approved):
Feel free to merge once test passes; you can have a separate pull request that dumps properties out in planner test.

PR #67, opened 2024-02-14
feat: Eliminate Duplicated Expr Rule

Sweetsuro:
# Major Changes
Add an eliminate duplicate expression rule which:
1. Removes duplicate sort expressions
2. Removes duplicate aggregate group bys

Also, this PR adds derive traits for Hash, PartialEq, and Eq for RelNode.

Examples:
`select * from t1 order by id, name, id desc, id asc, name desc` becomes 
`select * from t1 order by id, name`

`select * from t1 group by id, name, id` becomes 
`select * from t1 group by id, name`

## Rule Type
Heuristics (always apply), Transformation Rule (logical to logical)

### DIFF
--- a/optd-datafusion-repr/src/rules/eliminate_duplicated_expr.rs
+++ b/optd-datafusion-repr/src/rules/eliminate_duplicated_expr.rs
@@ -1,15 +1,81 @@
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
+use std::sync::Arc;
 
+use itertools::Itertools;
 use optd_core::rules::{Rule, RuleMatcher};
 use optd_core::{optimizer::Optimizer, rel_node::RelNode};
 
-use crate::plan_nodes::{ConstantType, LogicalEmptyRelation, OptRelNode, OptRelNodeTyp};
+use crate::plan_nodes::{
+    Expr, ExprList, LogicalSort, OptRelNode, OptRelNodeTyp, PlanNode, SortOrderExpr, SortOrderType,
+};
 
 use super::macros::define_rule;
 
 define_rule!(
     EliminateDuplicatedSortExprRule,
     apply_eliminate_duplicated_sort_expr,
-    (Sort, child, [cond])
+    (Sort, child, [exprs])
 );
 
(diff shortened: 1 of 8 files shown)


optd-datafusion-repr/src/rules/eliminate_duplicated_expr.rs
```diff
@@ -0,0 +1,117 @@
...
+        );
+        return vec![node.into_rel_node().as_ref().clone()];
+    }
+    vec![]
+}
+
+define_rule!(
+    EliminateDuplicatedAggExprRule,
+    apply_eliminate_duplicated_agg_expr,
+    (Agg, child, [exprs], [groups])
```
skyzh: no need to capture full bindings of exprs, just  use `exprs` would be fine. Need to add `ExprList::from_group` and `Expr::from_group` after changing it to a group capture.

skyzh (approved):
Rest LGTM

skyzh:
and adding a second rule that applies on the same expression might cause panic, at that time you should move the rule into the heuristics phase (need to create the heuristic optimizer in df-optimizer)

optd-datafusion-repr/src/rules/eliminate_duplicated_expr.rs
Sweetsuro: Dumb question, what's the difference between using [exprs] and exprs? I it's guess all expression bindings vs matching on one group...?

skyzh: > Note that variables like a, b, c will only match the group ID of a child, while [cond1] will expand the group IDs and retrieve the concrete expressions. [xxx] is useful when matching the SQL expression child.

https://cmu-db.github.io/optd/rule_engine.html

Sweetsuro: Fixed. Also added an implementation for `ExprList::from_group` since it didn't exist.

skyzh:
feel free to merge if you feel it's ready :)

PR #69, opened 2024-02-16
feat: support tpch q5 and q6

Gun9niR:
- Support gt and lt for bin op type
- Support `LikeExpr` (it's for Q2, but it turns out Q2 requires more work)
- Support for cast types

### DIFF
--- a/optd-datafusion-bridge/src/into_optd.rs
+++ b/optd-datafusion-bridge/src/into_optd.rs
@@ -106,7 +106,7 @@ impl OptdPlanContext<'_> {
                 }
                 ScalarValue::Float64(x) => {
                     let x = x.as_ref().unwrap();
-                    Ok(ConstantExpr::decimal(*x).into_expr())
+                    Ok(ConstantExpr::float64(*x).into_expr())
                 }
                 ScalarValue::Utf8(x) => {
                     let x = x.as_ref().unwrap();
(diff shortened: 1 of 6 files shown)


optd-datafusion-bridge/src/into_optd.rs
```diff
@@ -101,6 +104,10 @@ impl OptdPlanContext<'_> {
                     let x = x.as_ref().unwrap();
                     Ok(ConstantExpr::int64(*x).into_expr())
                 }
+                ScalarValue::Float64(x) => {
+                    let x = x.as_ref().unwrap();
```
skyzh, line 108: F64 is different from decimal.

optd-sqlplannertest/tests/tpch.planner.sql
```diff
@@ -88,6 +88,211 @@ CREATE TABLE LINEITEM (
...
+    n_name AS nation,
+    SUM(l_extendedprice * (1 - l_discount)) AS revenue
+FROM
+    customer,
+    orders,
+    lineitem,
+    supplier,
+    nation,
+    region
+WHERE
```
skyzh, line 102: I feel the other team should prioritize filter push down rules otherwise join executors never get sufficient tests.

optd-datafusion-bridge/src/into_optd.rs
Gun9niR: Added `Float64` type.

optd-sqlplannertest/tests/tpch.planner.sql
Gun9niR: The goal for this PR is just to ensure the queries do not crash.

AveryQi115 (approved):
LGTM! We'll quickly add filter push down rules this weekend or early next week

jurplel (approved):
nice work!

PR #70, opened 2024-02-16
feat: "hello world" selectivity computation

patrickhyw:
**Description**: reason this PR just contains the simplest possible selectivity computation is because computing selectivity at all requires lots of refactoring, so I want that refactoring reviewed separately from the "core" selectivity computation logic.

**Demo**: passing `test_colref_eq_constint_in_mcv` test
Screenshot 2024-02-16 at 12 37 56


optd-datafusion-repr/src/cost/base_cost.rs
```diff
@@ -115,7 +131,35 @@ impl CostModel<OptRelNodeTyp> for OptCostModel {
...
+                let selectivity = match context {
+                    Some(context) => {
+                        if let Some(optimizer) = optimizer {
+                            let column_refs = optimizer
+                                .get_property_by_group::<ColumnRefPropertyBuilder>(
+                                    context.group_id,
+                                    1,
+                                );
+                            let expr_group_id = context.children_group_ids[1];
+                            let expr_trees = optimizer.get_all_group_bindings(expr_group_id, false);
```
Gun9niR, line 143: Should we check `expr_trees.len() == 1` in case there's some bug?

patrickhyw: I already do right? cuz first() returns an Option that I unwrap safely

Gun9niR: What if there are multiple trees due to some bug?

optd-datafusion-repr/src/cost/base_cost.rs
```diff
@@ -192,7 +236,168 @@ impl CostModel<OptRelNodeTyp> for OptCostModel {
...
+                    let right_child = expr_tree.child(1);
+
+                    if left_child.as_ref().typ == OptRelNodeTyp::ColumnRef {
+                        let col_ref_idx = left_child.as_ref().data.as_ref().unwrap().as_u64();
+                        // this is always safe since col_ref_idx was initially a usize in ColumnRefExpr::new()
+                        let usize_col_ref_idx = col_ref_idx as usize;
+                        if let ColumnRef::BaseTableColumnRef { table, col_idx } =
+                            &column_refs[usize_col_ref_idx]
+                        {
+                            if let OptRelNodeTyp::Constant(_) = right_child.as_ref().typ {
```
Gun9niR, line 275: Do we need to check the case where `left_child` is `Constant` and `right_child` is `ColumnRef`?

patrickhyw: resolved on discord: that's a future PR

PR #71, opened 2024-02-17
feat: add tdigest data structure for statistics

AlSchlo:
This pull request implements a simplified version of [Ted Dunning's TDigest algorithm](https://arxiv.org/pdf/1902.04023.pdf) for efficient quantile/cdf computations.

It supports a fully parallelizable, memory-bounded computation scheme, along with an easy API.

Simplified for two reasons:
- The linear interpolation is only done for the quantile, not the cdf.
- Linear interpolation in general could be done more efficiently if we leverage unit-weighted centroids at the edges, as described in the paper.
 
Both are marked as TODOs in the code for future work.

Most open-source implementations found online had bugs, were incomplete, or were too complex (i.e., **poorly written**).

Includes unit tests on uniform and weighted distributions.

Next steps: Implementing the HyperLogLog algorithm for NDistinct.


gungnir/src/stats/tdigest.rs
```diff
@@ -0,0 +1,335 @@
...
+//! Ted Dunning's paper:
+//! "Computing Extremely Accurate Quantiles Using t-Digests" (2019).
+//! For more details, refer to: https://arxiv.org/pdf/1902.04023.pdf
+
+use itertools::Itertools;
+use std::f64::consts::PI;
+
+// The TDigest structure for the statistical aggregator to query quantiles.
+pub struct TDigest {
+    centroids: Vec<Centroid>, // A sorted array of Centroids, according to their mean.
```
yliang412: You could use doc comments i.e. `///` above the structs and fields so the comments show up in the documentation. For example,

```rust
/// The TDigest structure for the statistical aggregator to query quantiles.
pub struct TDigest {
    /// A sorted array of Centroids, according to their mean.
    centroids: Vec<Centroid>, 
...
}

File: gungnir/src/stats/tdigest.rs [hunk 2]
```diff
@@ -0,0 +1,335 @@
...
+            None => {
+                return self;
+            }
+        };
+
+        // Iterate over ordered and merged Centroids (starting from index 1).
+        for centroid in sorted_centroids {
+            let q_new = (tmp_centroid.weight + centroid.weight) as f64 / total_weight as f64;
+            if (q_curr + q_new) <= q_limit {
+                tmp_centroid.merge(&centroid)
```
yliang412:
```suggestion
                tmp_centroid.merge(centroid)
```

yliang412 (commented):
I haven't read the paper in 100% detail, but overall LGTM! 

Could we keep the license consistent with the one for the entire repository ([LICENSE](https://github.com/cmu-db/optd/blob/main/LICENSE))? I also suggest that you could add a short description for the `gungnir` crate under the structure section in the repo [README](https://github.com/cmu-db/optd?tab=readme-ov-file#structure).

gungnir/src/stats/tdigest.rs
AlSchlo: Where did I change the License? I don't think I added anything about the license.
Do you mean the cool header at the top of each Gungnir™ file? :joy:

yliang412: Oh yea. I was talking about those. Those are cool though lol.

PR #73, opened 2024-02-17
feat: support type expr

Gun9niR:
`DataTypeExpr` wraps `arrow_schema::DataType`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2498,6 +2498,7 @@ name = "optd-datafusion-repr"
 version = "0.1.0"
 dependencies = [
  "anyhow",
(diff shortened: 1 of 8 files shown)


yliang412 (approved):
LGTM. We do still have the Decimal128 precision lost for `Value` right? Since that part is not part of a cast.

skyzh:
I think it’s stored in DF type information? Need double check if there are any other places we are hard-coding precision.

Gun9niR:
Precision is stored in DF type. The problem is decimal is stored as `Value::Float`. It's not relevant to `Cast`.

Issue #68, opened 2024-02-15
Tracking: make sure optd does not crash for TPC-H queries


Gun9niR:
- [x] Q1 #78 
  - 🚧 Not working in perftest: `Error: ArrowError(NotYetImplemented("Unsupported Interval Expression with value \"115 day (3) seconds\""))`
- [x] Q2 #80 
- [x] Q3 #86 
- [ ] Q4
  - optd panics with `cannot find best binding for group` (Not supporting left semi join)
- [x] Q5 #69 
- [x] Q6 #69 
  - 🚧 Working in optd but not in DataFusion: `Error: Arrow error: Invalid argument error: Invalid comparison operation: Decimal128(15, 2) >= Decimal128(20, 0)`
- [x] Q7 #72 
- [x] Q8 https://github.com/cmu-db/optd/pull/64
- [x] Q9 #72 
- [x] Q10 #78 
- [x] Q11
  - Non-equal join filer
- [x] Q12 #77 
- [x] Q13
  - DF `Join` `filter` says `Filters applied during join (non-equi conditions)`, but there can actually be equal comparisons!
  - Non-equal join filter
- [x] Q14 #78 
- [x] Q15 #86 
  - 🚧 Not working in perftest: `unsupported plan node: CreateView: Bare { table: "revenue0" }`
- [ ] Q16
  - optd panics with `cannot find best binding for group`(Not supporting left anti join)
- [x] Q17 #86 
- [ ] Q18
  - optd panics with `cannot find best binding for group` (Not supporting left semi join)
- [x] Q19 #77 
- [ ] Q20
  - optd panics with `cannot find best binding for group` (Not supporting left semi join)
- [ ] Q21
  - optd panics with `cannot find best binding for group` (Not supporting left semi and left anti join)
- [ ] Q22
  - optd panics with `cannot find best binding for group`(Not supporting left anti join)
### Missing features
- [x] Subquery: Q2, Q3, Q4, Q11, Q16, Q17, Q20, Q21, Q22
  - Correlated subquery is not supported in DF physical optimizer. DF's logical optimizer eliminates `OuterReferenceColumn` by moving the column to the outer query. Since we do not know how to access columns of the outer query, I think we should enable DF's logical optimizer when dealing with subqueries.
  - ~With DF's logical optimizer enabled, we get `Error during planning: table 'datafusion.public.xxxx' not found` error.~ When using `explain_with_logical` in the planner test, make sure to execute the DDLs with `execute_with_logical`!
[...]

PR #76, opened 2024-02-19
test: add planner test for working tpch query (Q18)

Gun9niR:
For now only Q5-9 and Q18 works. I've listed the missing features in #68.


yliang412 (approved):
Thanks for keeping track of the missing feature.

PR #77, opened 2024-02-19
feat: support InListExpr

Gun9niR:
This enables tpch Q12 and Q19.

### DIFF
--- a/optd-datafusion-repr/src/plan_nodes/expr.rs
+++ b/optd-datafusion-repr/src/plan_nodes/expr.rs
@@ -790,6 +790,64 @@ impl OptRelNode for LikeExpr {
             vec![
                 ("expr", self.child().explain()),
                 ("pattern", self.pattern().explain()),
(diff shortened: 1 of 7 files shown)


optd-datafusion-repr/src/plan_nodes/expr.rs
```diff
@@ -790,6 +790,64 @@ impl OptRelNode for LikeExpr {
...
+
+    pub fn child(&self) -> Expr {
+        Expr(self.0.child(0))
+    }
+
+    pub fn list(&self) -> ExprList {
+        ExprList::from_rel_node(self.0.child(1)).unwrap()
+    }
+
+    pub fn negated(&self) -> bool {
```
yliang412, line 818: It's not clear what the negated means. Could we add some docs?

yliang412 (approved):
LGTM. I would suggest documenting how we are storing things + what helpers do so other people don't need to go into datafusion doc to search for it.

PR #80, opened 2024-02-20
test: add tpc-h q2

skyzh:
This pull request fixes the context for testing and uses a single catalog for both DF logical + optd logical optimizer.

* With DF logical optimizer enabled, it would take a really long time for optd to search the plan space and it always hit the budget. Need to properly implement pruning and later remove the partial explore budget from optd.
* The plan may be unstable across different platform as access to HashMap is not deterministic.


yliang412 (approved):
LGTM

PR #81, opened 2024-02-20
feat: core filter selectivity

patrickhyw:
"Core" means I am handling all cases where we don't fall back to hardcoded defaults (e.g. missing statistics, "col = col", "col = subquery", etc.).

Note the difference between this PR and [hello world selectivity](https://github.com/cmu-db/optd/pull/70). In this PR, the only file that changed is `base_cost.rs`, because all the "infrastructure" was already set up in "hello world selectivity"

I have written 20 unit tests to test this core logic in a variety of scenarios (nulls vs no nulls, value in MCVs vs not in MCVs, reversing the order of children, etc.) I chose to write code-based unit tests instead of SQLPlanner-like tests because it allows fine-grained control of the expression tree down to the order of children (and because it takes a lot more work to modify SQLPlanner to output cardinality). I made an effort to make the unit tests less brittle by defining helper functions such as `bin_op()` or `const_i32()` for constructing expression trees. If our expression tree representation changes in the future, it is likely that only these helper functions need to be changed to make the unit tests work again.

Some TPC-H queries require "non-core" logic, so this code doesn't currently run with all of TPC-H. To avoid crashing, I simply return INVALID_SELECTIVITY (0.001) for any branches that aren't a part of the "core". I will handle "non-core" logic in a future PR.

Another case not being handled is comparisons (</<=/>=/>) between `Value` objects. Handling this requires further discussion because not all Values have a meaningful "order". In the meantime, I hardcoded converting all `Value` objects to `i32` to perform comparisons.


optd-datafusion-repr/src/cost/base_cost.rs
```diff
@@ -261,55 +287,287 @@ impl OptCostModel {
...
+        &self,
+        bin_op_typ: BinOpType,
+        left: OptRelNodeRef,
+        right: OptRelNodeRef,
+        column_refs: &GroupColumnRefs,
+    ) -> f64 {
+        assert!(bin_op_typ.is_comparison());
+
+        // the # of column refs determines how we handle the logic
+        let mut col_ref_nodes = vec![];
```
Gun9niR, line 339: Can we use `ColumnRefExpr::from_rel_node` to convert `left` and `right` here? It's safer and you don't need to check if data is `None` below by just calling `ColumnRefExpr::index`.

optd-datafusion-repr/src/cost/base_cost.rs
```diff
@@ -261,55 +287,287 @@ impl OptCostModel {
...
+                            if is_left_col_ref {
+                                self.get_column_range_selectivity(
+                                    table, *col_idx, value, true, false,
                                 )
                             } else {
-                                INVALID_SELECTIVITY
+                                self.get_column_range_selectivity(
+                                    table, *col_idx, value, false, false,
+                                )
                             }
```
Gun9niR: Can change to

```rust
self.get_column_range_selectivity(table, *col_idx, value, is_left_col_ref, false)
```

optd-datafusion-repr/src/cost/base_cost.rs
```diff
@@ -261,55 +287,287 @@ impl OptCostModel {
...
+                            if is_left_col_ref {
+                                self.get_column_range_selectivity(
+                                    table, *col_idx, value, false, false,
+                                )
+                            } else {
+                                self.get_column_range_selectivity(
+                                    table, *col_idx, value, true, false,
+                                )
+                            }
+                        }
```
Gun9niR: ditto

optd-datafusion-repr/src/cost/base_cost.rs
```diff
@@ -261,55 +287,287 @@ impl OptCostModel {
...
+        is_col_eq_val: bool,
+    ) -> Option<f64> {
+        if let Some(per_table_stats) = self.per_table_stats_map.get(table) {
+            if let Some(per_column_stats) = per_table_stats.per_column_stats_vec.get(col_idx) {
+                // because distr does not include the values in MCVs, we need to compute the CDFs there as well
+                // because nulls return false in any comparison, they are never included when computing range selectivity
+                let distr_leq_freq = per_column_stats.distr.cdf(value);
+                let value_clone = value.clone(); // clone the value so that we can move it into the closure to avoid lifetime issues
+                                                 // TODO: in a future PR, figure out how to make Values comparable. rn I just hardcoded as_i32() to work around this
+                let pred = Box::new(move |val: &Value| val.as_i32() <= value_clone.as_i32());
```
Gun9niR, line 495:
```suggestion
                let const_value = value.as_i32(); // TODO: in a future PR, figure out how to make Values comparable. rn I just hardcoded as_i32() to work around this
                let pred = Box::new(move |val: &Value| val.as_i32() <= const_value);
```

optd-datafusion-repr/src/cost/base_cost.rs
```diff
@@ -345,61 +613,695 @@ mod tests {
...
+        })
+    }
+
+    fn un_op(op_type: UnOpType, child: OptRelNodeRef) -> OptRelNodeRef {
+        Arc::new(RelNode::<OptRelNodeTyp> {
+            typ: OptRelNodeTyp::UnOp(op_type),
+            children: vec![child],
+            data: None,
+        })
+    }
```
Gun9niR: Can we create the `XxxExpr` first and then use `into_rel_node`? This way we don't need to worry about repr change in the core.

optd-datafusion-repr/src/cost/base_cost.rs
```diff
@@ -345,61 +613,695 @@ mod tests {
...
+        plan_nodes::{BinOpType, ConstantType, OptRelNodeRef, OptRelNodeTyp, UnOpType},
         properties::column_ref::ColumnRef,
     };
 
-    use super::{MostCommonValues, OptCostModel, PerColumnStats, PerTableStats};
+    use super::{Distribution, MostCommonValues, OptCostModel, PerColumnStats, PerTableStats};
 
     struct MockMostCommonValues {
         mcvs: HashMap<Value, f64>,
     }
```
Gun9niR, line 608: You can define a `new` and `empty` method on it to avoid calling `.iter().collect()` every time.

optd-datafusion-repr/src/cost/base_cost.rs
```diff
@@ -345,61 +613,695 @@ mod tests {
...
 
+    struct MockDistribution {
+        cdfs: HashMap<Value, f64>,
+    }
```
Gun9niR, line 612: ditto

optd-datafusion-repr/src/cost/base_cost.rs
```diff
@@ -345,61 +613,695 @@ mod tests {
...
+
+        fn total_freq(&self) -> f64 {
+            self.mcvs.values().sum()
+        }
+
+        fn freq_over_pred(&self, pred: Box<dyn Fn(&Value) -> bool>) -> f64 {
+            self.mcvs
+                .iter()
+                .filter(|(val, _)| pred(val))
+                .map(|(_, freq)| freq)
```
Gun9niR, line 639:
```suggestion
                .values()
```

Gun9niR (approved):
The logic LGTM!

optd-datafusion-repr/src/cost/base_cost.rs
patrickhyw: I wanted a solution that would work even after we start passing `Value` instead of `i32`

optd-datafusion-repr/src/cost/base_cost.rs
patrickhyw: fixed

optd-datafusion-repr/src/cost/base_cost.rs
patrickhyw: fixed

optd-datafusion-repr/src/cost/base_cost.rs
patrickhyw: fixed

optd-datafusion-repr/src/cost/base_cost.rs
patrickhyw: I don't think the .filter() object has .values()

Gun9niR: Right, my bad

PR #86, opened 2024-02-24
test: add tpch 3, 15, 17 planner test

Gun9niR:
After #80, we can test subqueries with DF logical optimizer enabled. The reasons not all queries are working are:

- Multiple join filters, possibly containing non-eq predicate.
- optd panicking with `cannot find best binding for group`. I suspect this is due to the partial explore?

In addition, I realized I pasted the wrong query for q18 🥵 (that made me double-checked whether other queries are correct), so it is removed now.


skyzh:
The logical plan optd gets from DF should be directly convertible to physical plan. Therefore, it’s more likely to be a bug with optd if it cannot produce a plan.

skyzh:
Otherwise feel free to merge, I don’t plan to do a review.

PR #88, opened 2024-03-01
feat: add RuleWrapper and make the optimizer a hybrid optimizer

AveryQi115:
# Design
This pr is for adding a rule wrapper and hybrid logic to make cascades optimizer support two types of logic rules: rules registered as heuristics and rules registered as cascades.
- heuristic rule, which should always replace the old expr using the new expr if rule is applied
- heuristic rule should only return 1 or 0 expr, if not match, return the 0 expr
- heuristic rule's newly generated expr would be an expr not explored yet, only marks the heuristic rule which generates it as fired
- heuristic rule's newly generated expr might have been existed in memo, in that case two groups should be merged and old expr should be marked as a dead end (all rules fired for it)

# Major Changes
- Add RuleWrapper and optimizeType
- Change OptimizeExpression Task and ApplyRuleTask logic
- Add replace_group_expr for memo and optimizer

# why we are using a hybrid optimizer instead of two stage optimizers including heuristics and cascades?
Currently the columnRef property define emptyRelation derive rules as derive instead of a real column ref. However, this is not true for a two-stage optimizer. For instance, if we have a query like this select * from t1 join t2 on false, the original plan tree is Projection(columnRef 0,1,2,3) -> InnerJoin -> TableScan, after heuristic rules it becomes Projection(columnRef 0,1,2,3) -> EmptyRelation. But the emptyRelation has no connected table name or properties containing its schema(even if we create infer properties rules in heuristics, CascadesOptimizer cannot get the properties generated by heuristicOptimizers), but the projection node require columnref 0,1,2,3, so the columnRef derive rule will panic as it cannot infer from the children of emptyRelation. If we put heuristic rules in cascades optimizer, it works fine as the original table scan is stored in original plan and can be fetched from catalog, after transformation, the same group share the same properties, so it works fine.

If we're going to create a two-stage optimizer, it is hard to share properties/schema across two optimizers. No matter we're modifying current heuristic implementations with infer properties or creating a heuristic optimizer from cascades framework with customized task. (because the schema are either stored in memo or from catalog with table name, it cannot be easily shared with outside calls)

# future tasks
the lack of maintaining properties in remembered state might break the adaptive design as the emptyRelation is record and no connected schema without inferred properties.

### DIFF
--- a/optd-core/src/cascades/memo.rs
+++ b/optd-core/src/cascades/memo.rs
@@ -248,12 +248,15 @@ impl<T: RelNodeTyp> Memo<T> {
         self.groups.insert(group_id, group);
     }
 
+    // return true: replace success, the expr_id is replaced by the new rel_node
+    // return false: replace failed as the new rel node already exists in other groups,
+    //             the old expr_id should be marked as all rules are fired for it
     pub fn replace_group_expr(
         &mut self,
         expr_id: ExprId,
         replace_group_id: GroupId,
         rel_node: RelNodeRef<T>,
-    ) {
+    ) -> bool {
         let replace_group_id = self.get_reduced_group_id(replace_group_id);
 
         if let Entry::Occupied(mut entry) = self.groups.entry(replace_group_id) {
@@ -286,15 +289,16 @@ impl<T: RelNodeTyp> Memo<T> {
                 let group_id = self.get_reduced_group_id(group_id);
                 self.merge_group_inner(replace_group_id, group_id);
 
-                // TODO: how to safely remove the old expr from the group?
-                return;
+                // TODO: instead of remove this expr from the old group,
+                // we mark the expr as all rules have been fired to make it a dead end
+                return false;
             }
 
             self.expr_id_to_expr_node
                 .insert(expr_id, memo_node.clone().into());
             self.expr_node_to_expr_id.insert(memo_node.clone(), expr_id);
 
-            return;
+            return true;
         }
         unreachable!("group not found in replace_group_expr");
     }
--- a/optd-core/src/cascades/memo.rs
+++ b/optd-core/src/cascades/memo.rs
@@ -248,6 +248,57 @@ impl<T: RelNodeTyp> Memo<T> {
         self.groups.insert(group_id, group);
     }
 
+    pub fn replace_group_expr(
+        &mut self,
+        expr_id: ExprId,
+        replace_group_id: GroupId,
+        rel_node: RelNodeRef<T>,
+    ) {
+        let replace_group_id = self.get_reduced_group_id(replace_group_id);
+
+        if let Entry::Occupied(mut entry) = self.groups.entry(replace_group_id) {
+            let group = entry.get_mut();
+            if !group.group_exprs.contains(&expr_id) {
+                unreachable!("expr not found in group in replace_group_expr");
+            }
+
+            let children_group_ids = rel_node
+                .children
+                .iter()
+                .map(|child| {
+                    if let Some(group) = child.typ.extract_group() {
+                        group
+                    } else {
+                        self.add_new_group_expr(child.clone(), None).0
+                    }
+                })
+                .collect::<Vec<_>>();
+
+            let memo_node = RelMemoNode {
+                typ: rel_node.typ.clone(),
+                children: children_group_ids,
+                data: rel_node.data.clone(),
+            };
+
+            // if the new expr already in the memo table, merge the group and remove old expr
+            if let Some(&expr_id) = self.expr_node_to_expr_id.get(&memo_node) {
+                let group_id = self.get_group_id_of_expr_id(expr_id);
+                let group_id = self.get_reduced_group_id(group_id);
+                self.merge_group_inner(replace_group_id, group_id);
+
+                // TODO: how to safely remove the old expr from the group?
+                return;
+            }
+
+            self.expr_id_to_expr_node
+                .insert(expr_id, memo_node.clone().into());
+            self.expr_node_to_expr_id.insert(memo_node.clone(), expr_id);
+
+            return;
+        }
+        unreachable!("group not found in replace_group_expr");
+    }
+
     fn add_new_group_expr_inner(
         &mut self,
         rel_node: RelNodeRef<T>,
--- a/optd-core/src/cascades/tasks/apply_rule.rs
+++ b/optd-core/src/cascades/tasks/apply_rule.rs
@@ -12,7 +12,7 @@ use crate::{
         GroupId,
     },
     rel_node::{RelNode, RelNodeTyp},
-    rules::RuleMatcher,
+    rules::{RuleMatcher, RuleType},
 };
 
 use super::Task;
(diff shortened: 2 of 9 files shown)


AveryQi115:
It's a very basic design for making heuristic based on cascades. Looking for advice and suggestions.

skyzh:
I don't think we should have norm rules. All rules should work in both heuristics and cascades optimizer. Norm rules are simply rules that are better running in heuristics instead of cascades.

skyzh:
And therefore, to make cascades into heuristics, we should change the tasks being run in the optimizer. There could be a HeuristicsApplyRule task that does not maintain the equality of expressions in a group.

AveryQi115:
> And therefore, to make cascades into heuristics, we should change the tasks being run in the optimizer. There could be a HeuristicsApplyRule task that does not maintain the equality of expressions in a group.

You mean still maintaining two optimizer: one is cascades, one is heuristics. And to change the task in heuristics? or adding a new task in cascades optimizer? 

I think the current design is somewhat similar? It does not maintain the equality of expressions, just directly replace the RelMemoNode of certain expr.

skyzh:
Yep, but I’m not in favor of having a new type of rule.

AveryQi115:
I'm still a bit confused.

1. If we're going to add a new heuristic optimizer out of the cascades codes and change applyTaskRule, there will be a lot of duplicated codes and there's actually no need to write heuristic in cascades framework way.

2. If we just add a task like applyHeuristicRuleTask, it still doesn't solve the problem. Because transform rules and heuristic rules are similar but not the same. For example, transform rule will generate vec![] if the expr doesn't match inner conditions, but heuristic rule will return original expr, and transform rule can return multiple exprs but heuristic mostly only return one.

skyzh:
Broadly speaking, a rule should always be a matcher + a transform function that takes the matching result and produce a new plan node. Therefore, any rule should be able to do transformations in both cascades and heuristics, and it is uncommon to have a special rule type (it sounds like a rule property) to do that thing.

One example is filter push-down rule. We know it's always good to apply it, and that's why people would usually put it in the heuristics phase of the optimization, though it is possible to put it in the cascades and let the optimizer figure out whether to apply it or not.

Another example is expression simplify rule. It should also work in both cascades and heuristics in the future, though our optimizer framework does not support it.

Back to your question,

> If we're going to add a new heuristic optimizer out of the cascades codes and change applyTaskRule, there will be a lot of duplicated codes and there's actually no need to write heuristic in cascades framework way.

I feel the only thing that we will need to implement is a for loop that applies all rules one by one. We can reuse the rule matcher and binding generator in the cascades. The apply rule function is very short: https://github.com/cmu-db/optd/blob/61380184e21f4836dc919cc22539005fee517344/optd-core/src/heuristics/optimizer.rs#L132-L158

> If we just add a task like applyHeuristicRuleTask, it still doesn't solve the problem. Because transform rules and heuristic rules are similar but not the same. For example, transform rule will generate vec![] if the expr doesn't match inner conditions, but heuristic rule will return original expr, and transform rule can return multiple exprs but heuristic mostly only return one.


The two things seem *only* different on the interface level. You can define the interface to be "return the original expr if not matched", or "return nothing if it doesn't match". Both interfaces work in either heuristics/cascades.

AveryQi115:
> Broadly speaking, a rule should always be a matcher + a transform function that takes the matching result and produce a new plan node. Therefore, any rule should be able to do transformations in both cascades and heuristics, and it is uncommon to have a special rule type (it sounds like a rule property) to do that thing.
> 
> One example is filter push-down rule. We know it's always good to apply it, and that's why people would usually put it in the heuristics phase of the optimization, though it is possible to put it in the cascades and let the optimizer figure out whether to apply it or not.
> 
> Another example is expression simplify rule. It should also work in both cascades and heuristics in the future, though our optimizer framework does not support it.
> 
> Back to your question,
> 
> > If we're going to add a new heuristic optimizer out of the cascades codes and change applyTaskRule, there will be a lot of duplicated codes and there's actually no need to write heuristic in cascades framework way.
> 
> I feel the only thing that we will need to implement is a for loop that applies all rules one by one. We can reuse the rule matcher and binding generator in the cascades. The apply rule function is very short:
> 
> https://github.com/cmu-db/optd/blob/61380184e21f4836dc919cc22539005fee517344/optd-core/src/heuristics/optimizer.rs#L132-L158
> 
> > If we just add a task like applyHeuristicRuleTask, it still doesn't solve the problem. Because transform rules and heuristic rules are similar but not the same. For example, transform rule will generate vec![] if the expr doesn't match inner conditions, but heuristic rule will return original expr, and transform rule can return multiple exprs but heuristic mostly only return one.
> 
> The two things seem _only_ different on the interface level. You can define the interface to be "return the original expr if not matched", or "return nothing if it doesn't match". Both interfaces work in either heuristics/cascades.

thx for the insights and let me rethink the archi and I’ll update the pr.

optd-core/src/cascades/memo.rs
```diff
@@ -248,6 +248,61 @@ impl<T: RelNodeTyp> Memo<T> {
...
+                .collect::<Vec<_>>();
+
+            let memo_node = RelMemoNode {
+                typ: rel_node.typ.clone(),
+                children: children_group_ids,
+                data: rel_node.data.clone(),
+            };
+
+            // if the new expr already in the memo table, merge the group and remove old expr
+            if let Some(&expr_id) = self.expr_node_to_expr_id.get(&memo_node) {
```
skyzh, line 287: assert there is only one expr in the group?

optd-core/src/cascades/tasks/apply_rule.rs
```diff
@@ -188,13 +188,60 @@ impl<T: RelNodeTyp> Task<T> for ApplyRuleTask {
...
+                assert!(
+                    !rule.is_impl_rule(),
+                    "impl rule registered should not be registered as heuristics"
+                );
+
+                if let Some(group_id_2) = typ.extract_group() {
+                    // If this is a group, merge the groups!
+                    optimizer.merge_group(group_id, group_id_2);
+                    // mark the old expr as a dead end
+                    (0..optimizer.rules().len())
```
skyzh, line 222: this looks like a costly operation, do we have to do that?

optd-datafusion-repr/src/lib.rs
```diff
@@ -107,17 +139,48 @@ impl DatafusionOptimizer {
...
+        rule_wrappers.insert(
+            2,
+            Arc::new(RuleWrapper::new(
+                Arc::new(ProjectionPullUpJoin::new()),
+                OptimizeType::Cascades,
+            )),
+        );
+        rule_wrappers.insert(
+            3,
+            Arc::new(RuleWrapper::new(
```
skyzh: consider defining something like `(dyn Rule).as_cascades()`, or `RuleWrapper::new_cascades`.

skyzh (approved):
Generally LGTM, please update the design doc later :)

optd-core/src/cascades/memo.rs
AveryQi115: No, in this case, the generated expr might be already in another group, then we merge these two groups and instead of removing old expr, we make it an expr which will not be applied with any rules.

optd-core/src/cascades/tasks/apply_rule.rs
AveryQi115: The heuristic rule replaces the old expr with new expr. But if the new expr is a placeholder of a group, we cannot simply replace the old RelMemoNode. Instead, we merge two groups.

And mark the old expr as a dead end is like a pruning method. I'm afraid that simply remove the old one might cause its parent node unable to find the child?

AveryQi115: I'll create an issue about the possible costly operation and how it can be improved and I'll merge the current pr first as currently with a small number of rules the situation is quite rare.

PR #91, opened 2024-03-06
feat: add misragries data structure for statistics

AlSchlo:
This pull request implements the Misra-Gries Summary data structure as described in [Cormode et al. paper](https://people.csail.mit.edu/rrw/6.045-2017/encalgs-mg.pdf).

We **refined** the algorithm to ensure that K elements will always be returned, as long as K <= N. As such, this data structure returns **all** elements with frequency f >= (n/k), and may include additional leftovers (i.e. best guesses).

It supports a fully parallelizable, memory-bounded MCV computation scheme, through an easy API.

Includes unit tests on i32s.

Next steps: Second sweep to gather the associated frequency of each MCV idendified by the Misra Gries Summary.


optd-gungnir/src/stats/misragries.rs
```diff
@@ -0,0 +1,202 @@
...
+            .frequencies
+            .iter()
+            .for_each(|(key, occ)| self.insert_element(key.clone(), *occ));
+    }
+
+    /// Returns all elements with frequency f >= (n/k),
+    /// and may include additional leftovers.
+    pub fn most_frequent_keys(&self) -> Vec<&T> {
+        self.frequencies.keys().collect_vec()
+    }
```
patrickhyw, line 92: Is it possible to know the actual frequency (as a number between 0 and 1) of each of the elements too?

patrickhyw (approved):
LGTM except the one comment

optd-gungnir/src/stats/misragries.rs
AlSchlo: No, that's why we need the second pass.

PR #93, opened 2024-03-07
feat: structure for cardinality testing

patrickhyw:
**Summary**: Defined the main interfaces (`CardtestRunner` and `CardtestRunnerDBHelper`) and structure (the `optd-perftest` package) for cardinality testing. Wrote function to compute the Q-error of multiple databases.


optd-perftest/src/main.rs
```diff
@@ -0,0 +1,17 @@
...
+mod cardtest;
+mod postgres;
+
+fn main() {
+    let databases: Vec<Box<dyn CardtestRunnerDBHelper>> = vec![
+        Box::new(block_on(PostgresDb::new()).unwrap()),
+        Box::new(block_on(DatafusionDb::new()).unwrap()),
+    ];
+    let cardtest_runner = block_on(CardtestRunner::new(databases)).unwrap();
+    let qerrors = block_on(cardtest_runner.eval_qerrors("")).unwrap();
```
Gun9niR, line 16: If you think writing `block_on` everywhere is too cumbersome, you can use `#[tokio::main]` to change main into an async function.

optd-perftest/src/cardtest.rs
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    async fn eval_true_card(&self, _sql: &str) -> anyhow::Result<usize> {
+        Ok(10)
+    }
+
+    async fn eval_est_card(&self, _sql: &str) -> anyhow::Result<usize> {
+        Ok(5)
+    }
+}
```
Gun9niR: Might be better to move trait implementations to separate files.

Gun9niR (approved):
Rest LGTM!

optd-perftest/src/main.rs
patrickhyw: thanks for the note. I changed this in the tpch-kit PR so I won't change it here

optd-perftest/src/cardtest.rs
patrickhyw: fixed

PR #95, opened 2024-03-09
feat: load tpch into postgres

patrickhyw:
**Summary**: Code for installing/initializing/running Postgres as well as loading TPC-H into Postgres.

**Demo**:

TPC-H tables and queries are already generated. In the first run, we load the TPC-H data into Postgres. In the second run, we just use the already loaded `pgdata/`.

https://github.com/cmu-db/optd/assets/20631215/a412185c-7c61-436f-9232-2b0d61863c43

**Details**:
* Postgres installation/initialization/running is all cached and only done if necessary
* Loading benchmarks is cached as well and is based on the configuration of the benchmark _only if the benchmark is read-only_
* As before, caching logic is atomic
* Benchmarks and benchmark configs are extensible, including the caching logic (via `get_strid()`)


optd-perftest/src/postgres_db.rs
```diff
@@ -1,12 +1,205 @@
...
+        // (re)start postgres
+        db.install_postgres().await?;
+        db.init_pgdata().await?;
+        db.start_postgres().await?;
+
+        Ok(db)
+    }
+
+    /// Installs an up-to-date version of Postgres using the OS's package manager
+    async fn install_postgres(&self) -> Result<()> {
```
yliang412, line 64: It might be possible to use `cargo-pgrx` to get a cross-platform solution.

yliang412: https://github.com/pgcentralfoundation/pgrx/blob/develop/cargo-pgrx/README.md

patrickhyw: Thanks for the suggestion. As discussed on Discord, due to PRs branching off of this one, I will look into pgrx in a future PR.

PR #96, opened 2024-03-10
feat: benchmarking the q-error of tpch on postgres

patrickhyw:
**Summary**: Now automatically benchmarking the Q-error of a subset of TPC-H on Postgres.

**Demo**:

https://github.com/cmu-db/optd/assets/20631215/7f90395e-3f2e-4a73-ae04-a5f06af58635

**Details**:
* I chose a small subset of TPC-H to avoid edge cases and just demonstrate that it's "working in general"
    * Some edge cases, to be fixed in future PRs, include: (a) performance issues due to us not yet caching true cardinality, (b) tpch-kit generating queries with multiple statements when our code only handles queries with one statement.
* Automatically creating the Postgres connection/client to allow query execution.
* Defined a deterministic order of execution with `get_sql_fpath_ordered_iter()`.
* Now generates queries to 22 `[num].sql` files instead of one giant file to enable Q-error evaluation.
* Improved the interface of `CardtestRunnerDBHelper`. Details are in the comment above the trait.


optd-perftest/src/postgres_db.rs
```diff
@@ -45,35 +49,43 @@ impl PostgresDb {
...
         db.init_pgdata().await?;
         db.start_postgres().await?;
+        db.connect_to_postgres().await?;
 
         Ok(db)
     }
 
     /// Installs an up-to-date version of Postgres using the OS's package manager
-    async fn install_postgres(&self) -> Result<()> {
+    async fn install_postgres(&self) -> anyhow::Result<()> {
```
Gun9niR, line 68: The user should install PosgreSQL himself. The reasons are:

- Maintaining the installation script for different OS's is extra work.
- Calling `brew update` and `brew upgrade` will very likely mess up the user's environment. For example, I like to run my postgres instance in docker

optd-perftest/src/postgres_db.rs
```diff
@@ -26,10 +29,11 @@ pub struct PostgresDb {
 
 /// Conventions I keep for methods of this class:
 ///   - Functions should be idempotent. For instance, start_postgres() should not fail if Postgres is already running
+///       - For instance, this is why "createdb" is _not_ a function
 ///   - Stop and start functions should be separate
 ///   - Setup should be done in build() unless it requires more information (like benchmark)
 impl PostgresDb {
-    pub async fn build(verbose: bool) -> Result<Self> {
+    pub async fn build(verbose: bool) -> anyhow::Result<Self> {
```
Gun9niR: Usually it's called `new` in Rust

optd-perftest/src/benchmark.rs
```diff
@@ -1,6 +1,7 @@
 use crate::tpch::TpchConfig;
 
 pub enum Benchmark {
+    #[allow(dead_code)]
     Test,
```
Gun9niR, line 5: What's this for?

optd-perftest/src/benchmark.rs
```diff
@@ -1,6 +1,7 @@
 use crate::tpch::TpchConfig;
 
 pub enum Benchmark {
```
Gun9niR, line 3: We can also add a `Sql` variant in future PRs so the user can supply input from a .sql file, which is more flexible.

optd-perftest/src/postgres_db.rs
```diff
@@ -116,77 +144,130 @@ impl PostgresDb {
     /// Start the Postgres process if it's not already started
     /// It will always be started using the pg_ctl binary installed with the package manager
     /// It will always be started on port 5432
-    async fn start_postgres(&self) -> Result<()> {
+    async fn start_postgres(&self) -> anyhow::Result<()> {
```
Gun9niR, line 130: Should let the users start or stop the instance for the same reasons as letting them installing it.

optd-perftest/src/benchmark.rs
```diff
@@ -9,7 +10,7 @@ impl Benchmark {
     pub fn get_strid(&self) -> String {
```
Gun9niR:
```suggestion
    pub fn get_name(&self) -> String {
```

At first glance I'm like: is "strid" a typo of "stride"? 🤣

optd-perftest/src/postgres_db.rs
```diff
@@ -92,22 +104,38 @@ impl PostgresDb {
...
+            if self.verbose {
+                println!("[end] initializing pgdata");
+            }
         } else {
             #[allow(clippy::collapsible_else_if)]
             if self.verbose {
-                println!("skipped running initdb");
+                println!("[skip] initializing pgdata");
             }
         }
```
Gun9niR: I think it's enough to just let the user specify the database name and report an error if the database already exists.

optd-perftest/src/tpch.rs
```diff
@@ -7,8 +7,9 @@ use std::fs::File;
 use std::io;
 use std::path::{Path, PathBuf};
 
-const TPCH_KIT_REPO_URL: &str = "git@github.com:gregrahn/tpch-kit.git";
+const TPCH_KIT_REPO_URL: &str = "git@github.com:lmwnshn/tpch-kit.git";
```
Gun9niR, line 10: Is it possible to use git submodule instead of cloning the repo every time?

optd-perftest/src/postgres_db.rs
```diff
@@ -45,35 +49,43 @@ impl PostgresDb {
...
 
         Ok(db)
     }
 
     /// Installs an up-to-date version of Postgres using the OS's package manager
-    async fn install_postgres(&self) -> Result<()> {
+    async fn install_postgres(&self) -> anyhow::Result<()> {
         match OS {
             "macos" => {
                 if self.verbose {
```
Gun9niR: The code is cluttered with all the `if verbose` statements. A better way to implement this is to use some logging crate, and equate `verbose == true` to a debugging level of, for instance, `DEBUG`.

optd-perftest/src/tpch.rs
```diff
@@ -43,15 +47,16 @@ pub struct TpchKit {
 impl TpchKit {
     pub fn build(verbose: bool) -> io::Result<Self> {
```
Gun9niR: Maybe rename to `new`?

optd-perftest/src/tpch.rs
```diff
@@ -43,15 +47,16 @@ pub struct TpchKit {
...
         // build paths, sometimes creating them if they don't exist
-        let curr_dpath = env::current_dir()?;
+        // we assume that this is being run in the base optd repo dir
+        let optd_repo_dpath = env::current_dir()?;
```
Gun9niR, line 49: There's a command to get repo root dir: `git rev-parse --show-toplevel`

Gun9niR (commented):
The workflow LGTM, but some implementation details might pose some usability problems. I cannot try the testing framework for now because `brew upgrade` will surely break my local dev env. These issues include:

- Installing Postgres within our testing framework.
- Initialize a new postgres cluster in local dir. If I'm running posgres in docker, the container might not have access to it. We should let user supply it or simply panic when we cannot create the testing database.

In addition, currently what database to create, what benchmark to run are hard-coded in `main`. Recompiling the binary every time we change it is not a good idea. Ultimately we want them to be specified from some configuration file, e.g. yaml just like the planner test.

optd-perftest/src/postgres_db.rs
patrickhyw: Sounds good I agree with this.

optd-perftest/src/postgres_db.rs
patrickhyw: I called it `build` per the suggestion of [RPL Ch12.3](https://rust-book.cs.brown.edu/ch12-03-improving-error-handling-and-modularity.html) since it can fail.

optd-perftest/src/benchmark.rs
patrickhyw: It's just a really simple workload (one SELECT * statement) for testing. None of the code currently uses it but I think it's useful to leave in there.

optd-perftest/src/benchmark.rs
patrickhyw: I agree. This should be a future PR though as it has some complications: you need a file describing how data is loaded as well (or you can say the user has to load the data beforehand), you may need to make it a script that generates `.sql` files instead of just one `.sql` file to make it work with different databases, etc.

optd-perftest/src/benchmark.rs
patrickhyw: LOL strid is a weird name. I do want to keep the idea of an "id" in there though because it's meant to be unique. Maybe `get_stringid` or `get_semanticid`?

optd-perftest/src/tpch.rs
patrickhyw: Yeah, that's a good idea.

optd-perftest/src/postgres_db.rs
patrickhyw: Yup. makes sense.

optd-perftest/src/tpch.rs
patrickhyw: See my comment on the other `build` function.

optd-perftest/src/tpch.rs
patrickhyw: Yup, I handled this in a future PR.

optd-perftest/src/postgres_db.rs
patrickhyw: This is a bit trickier than installation. Right now, part of my code stops Postgres programmatically, deletes pgdata (this is safe because we use a custom pgdata dir), and then starts it programmatically again. More broadly, some actions (like configuration changes) require restarting the Postgres instance, so it's useful to be able to do that programmatically.

That said, there is a way to get rid of programmatically starting/stopping Postgres if necessary. Instead of deleting and recreating pgdata, we could just call `dropdb` and `createdb`. For actions that require restarting in the future, we can just crash and tell the user what manual actions they need to take.

optd-perftest/src/postgres_db.rs
patrickhyw: This gets into a question of whether the user or optd should manage "caching". I think it'd be a good idea for optd to do it to reduce what the user has to think about. I do agree that we shouldn't hardcode the database name as `"optd"` and instead name the database based on what workload is loaded there (like `"tpch_sf0.01"` for instance). Then, optd can choose to drop/create databases if necessary (like if the workload is not read-only).

optd-perftest/src/postgres_db.rs
Gun9niR: Yeah dropdb and createdb sounds better.

PR #103, opened 2024-03-12
feat: specify flags in a more flexible way in planner test

Gun9niR:
In planner test we might want to specify multiple flags for one task. For instance, for `explain` we might want `with_logical` and `verbose`. With the current syntax, however, we need to define multiple tasks for this: `explain_with_logical_verbose`, `explain_with_logical`, `explain_verbose`, ... This PR moves such options to a square bracket right after the task name, so where we used to have `explain_with_logical`/`explain_with_logical_verbose` we now have `explain[with_logical]`, `explain[with_logical, verbose]`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2696,6 +2696,7 @@ dependencies = [
  "dirs",
  "env_logger 0.9.3",
  "itertools",
+ "lazy_static",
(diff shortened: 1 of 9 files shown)


yliang412 (approved):
LGTM

PR #104, opened 2024-03-13
feat: display cost in explain

### LINKED ISSUE #89: Display cardinality and cost in `explain`
We should display cardinality and cost of the physical plan in `explain`.

We can implement this with `RelNodeMetaMap` introduced in #65, by storing the cost in `RelNodeMeta`. Since the `explain` output is produced in a separate traversal of the plan tree after optimization, and `Pretty` objects are immutable after construction, we should pass the `RelNodeMeta` as a parameter to `dispatch_explain`.

Gun9niR:
Closes #89. When using `explain verbose`, the cardinality of each plan node will be displayed.

Example:

```
❯ create table t1(v1 int);
0 rows in set. Query took 0.005 seconds.

Execution took 0.000 secs, Planning took 0.000 secs
❯ explain verbose select * from t1;

PhysicalProjection { exprs: [ #0 ], cost: weighted=1.06,row_cnt=1.00,compute=0.06,io=1.00 }
└── PhysicalScan { table: t1, cost: weighted=1.00,row_cnt=1.00,compute=0.00,io=1.00 }

❯ explain verbose select count(*) from t1;

PhysicalProjection { exprs: [ #0 ], cost: weighted=21.18,row_cnt=1.00,compute=20.18,io=1.00 }
└── PhysicalAgg
    ├── aggrs:Agg(Count)
    │   └── [ 1 ]
    ├── groups: []
    ├── cost: weighted=21.12,row_cnt=1.00,compute=20.12,io=1.00
    └── PhysicalScan { table: t1, cost: weighted=1.00,row_cnt=1.00,compute=0.00,io=1.00 }
```

### DIFF
--- a/optd-datafusion-repr/src/explain.rs
+++ b/optd-datafusion-repr/src/explain.rs
@@ -1,15 +1,25 @@
 use optd_core::rel_node::RelNodeMeta;
 use pretty_xmlish::Pretty;
 
-use crate::cost::ROW_COUNT;
+use crate::cost::{COMPUTE_COST, IO_COST, ROW_COUNT};
 
 pub trait Insertable<'a> {
     fn with_meta(self, meta: &RelNodeMeta) -> Self;
 }
 
 impl<'a> Insertable<'a> for Vec<(&'a str, Pretty<'a>)> {
+    // FIXME: this assumes we are using OptCostModel
     fn with_meta(mut self, meta: &RelNodeMeta) -> Self {
-        self.push(("rows", Pretty::display(&meta.cost.0[ROW_COUNT])));
+        self.push((
+            "cost",
+            Pretty::display(&format!(
+                "weighted={:.2},row_cnt={:.2},compute={:.2},io={:.2}",
+                meta.cost.0[0],
+                meta.cost.0[ROW_COUNT],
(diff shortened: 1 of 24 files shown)


optd-datafusion-repr/src/explain.rs
```diff
@@ -0,0 +1,15 @@
...
+
+use crate::cost::ROW_COUNT;
+
+pub trait Insertable<'a> {
+    fn with_meta(self, meta: &RelNodeMeta) -> Self;
+}
+
+impl<'a> Insertable<'a> for Vec<(&'a str, Pretty<'a>)> {
+    fn with_meta(mut self, meta: &RelNodeMeta) -> Self {
+        self.push(("rows", Pretty::display(&meta.cost.0[ROW_COUNT])));
```
yliang412: I would recommend maybe using `CostModel::explain`.

This would require to modify Cost model's signature from 
`fn explain(&self, cost: &Cost) -> String` to `fn explain(cost: &Cost) -> String`

```suggestion
        self.push(("cost", Pretty::display(CostModel::explain(&meta.cost))));
```

yliang412 (approved):
I made some comments on using the existing cost model's explain method to place into Insertable. Rest LGTM.

PR #105, opened 2024-03-13
feat: implement card test for df optimizer

Gun9niR:
Implement card test interfaces for Datafusion optimizer. After #104, the cost of each plan node can be displayed using `explain verbose`. Utilizing this feature, we can extract the estimated cardinality of the root plan node. The true cardinality is computed by counting the rows of the result set.

Also enable users to select queries to execute in perftest by specifying e.g. `--query-ids 2 3`

image


optd-perftest/src/main.rs
```diff
@@ -29,6 +30,9 @@ enum Commands {
         #[arg(long)]
         #[clap(default_value = "15721")]
         seed: i32,
+        #[arg(long)]
+        #[clap(value_delimiter = ' ', num_args = 1..)]
+        queries: Vec<u32>,
```
patrickhyw: I think `query_ids` is a better name. `queries` could mean a list of query objects, a list of query strings, etc.

optd-perftest/src/main.rs
```diff
@@ -37,12 +41,17 @@ async fn main() -> anyhow::Result<()> {
...
-    match &cli.command {
-        Commands::Cardtest { scale_factor, seed } => {
+    match cli.command {
+        Commands::Cardtest {
+            scale_factor,
+            seed,
+            queries,
+        } => {
             let tpch_config = TpchConfig {
                 database: String::from(TPCH_KIT_POSTGRES),
```
patrickhyw, line 64: Let's add a comment saying that we do this because DataFusion is Postgres-compatible, so people know it's not a mistake

optd-perftest/src/postgres_db.rs
```diff
@@ -327,8 +329,10 @@ impl PostgresDb {
 
     /// Extract the row count from a line of an EXPLAIN output
     fn extract_row_count(explain_line: &str) -> Option<usize> {
-        let re = Regex::new(r"rows=(\d+)").unwrap();
-        if let Some(caps) = re.captures(explain_line) {
+        lazy_static! {
+            static ref RE: Regex = Regex::new(r"row_cnt=(\d+\.\d+)").unwrap();
+        }
```
patrickhyw, line 238: Nice catch with the `lazy_static!`

optd-perftest/src/postgres_db.rs
```diff
@@ -327,8 +329,10 @@ impl PostgresDb {
 
     /// Extract the row count from a line of an EXPLAIN output
     fn extract_row_count(explain_line: &str) -> Option<usize> {
-        let re = Regex::new(r"rows=(\d+)").unwrap();
-        if let Some(caps) = re.captures(explain_line) {
+        lazy_static! {
+            static ref RE: Regex = Regex::new(r"row_cnt=(\d+\.\d+)").unwrap();
```
patrickhyw: `RE` is a vague name. How about `ROW_CNT_RE`?

optd-perftest/src/tpch.rs
```diff
@@ -235,8 +236,10 @@ impl TpchKit {
         tpch_config: &TpchConfig,
     ) -> io::Result<impl Iterator<Item = PathBuf>> {
         let this_genned_queries_dpath = self.get_this_genned_queries_dpath(tpch_config);
-        let query_i_iter = 3..=6; //NUM_TPCH_QUERIES; TODO(phw2): support all TPC-H queries in the future with custom tpch-kit + truecard caching
-        let sql_fpath_ordered_iter = query_i_iter
+        let sql_fpath_ordered_iter = tpch_config
+            .queries
+            .clone()
+            .into_iter()
             .map(move |query_i| this_genned_queries_dpath.join(format!("{}.sql", query_i)));
```
patrickhyw, line 249: Why not just use an implicit copy instead of an explicit clone here? `query_i` are just `u32` values

patrickhyw (commented):
LGTM other than small changes!

optd-perftest/src/tpch.rs
Gun9niR: Because if we use `tpch_config.query_ids.iter()`, the returned iter will capture a reference of `tpch_config`, which requires adding ugly lifetime markers to the signature. Cloning a small vector is not a lot of overhead, and I like to avoid lifetime markers when possible.

patrickhyw: Ah ok, makes sense!

patrickhyw (approved):

PR #106, opened 2024-03-13
feat: CLI for optd-perftest

patrickhyw:
**Summary**: An extensible CLI built with `clap` for the `optd-perftest` package

**Demo**:

https://github.com/cmu-db/optd/assets/20631215/05315dcc-6a91-4c4d-a9ea-fdb8d15ba675

**Details**:
* The CLI currently only exposes the `cardtest` option (cardinality testing) but can be extended to support more types of performance tests
* The `cardtest` option takes in scale factor and seed as options, but both have defaults as well
* Fixed a bug where we forgot to reconnect after reinitializing pgdata


Gun9niR (approved):
LGTM

PR #108, opened 2024-03-13
Feat: flatten filter and join cond to list of expr, simplify filter rule

AveryQi115:
# Major Changes
- BinOpExpr in Join cond and filter cond is changed to LogOpExpr
- LogOpExpr child is not a single ExprList but multiple exprs as children (which is different from previous logOpExpr)
- ColumnRef and Schema properties changed for LogOpExpr

# Design
Previously, LogOpExpr is designed to be `children: vec![ExprList]` which means the LogOpExpr only has one child of ExprList. And condition expressions in filter and join from datafusion will be converted to BinOpExpr, which means the LogOpExpr actually does not exist in optd. After converting conditions from datafusion to be LogOpExpr, the design of one child LogOpExpr will influence the ProjectPullUpJoinRule and other rules with schema recalculated as these rules simply traverse the expr's children in its relNode format, it will treat LogOpExpr children as only one child (ExprList), and when it further traverses ExprList it has no idea whether it is a LogOp list or a normal list.

So current design change the LogOpExpr to include multiple children to represent its real children size in RelNode format.
It has further impact on cost model and properties builder logic as previously List is the only type in Optd which contains multiple children.

#TODO
- [x] currently I simply copy properties derive and build rules for List as the rules for LogOpExpr, please check whether it is aligned with cost model's logic
- [x] the change break the selectivity infer rules and cost model calculation so it cannot find a best group bindings for several cases. Please update the selectivity calculation for logOpExpr.


optd-datafusion-repr/src/rules/filter.rs
```diff
@@ -0,0 +1,146 @@
...
+
+use super::macros::define_rule;
+
+define_rule!(
+    SimplifyFilterRule,
+    apply_simplify_filter,
+    (Filter, child, [cond])
+);
+
+fn simplify_log_expr(log_expr: OptRelNodeRef, changed: &mut bool) -> OptRelNodeRef {
```
patrickhyw, line 25: Could you add a comment describing what this function does? It seems pretty important but also complex

optd-datafusion-bridge/src/into_optd.rs
```diff
@@ -8,13 +8,30 @@ use datafusion_expr::Expr as DFExpr;
...
-    LogicalFilter, LogicalJoin, LogicalLimit, LogicalProjection, LogicalScan, LogicalSort,
-    OptRelNode, OptRelNodeRef, OptRelNodeTyp, PlanNode, SortOrderExpr, SortOrderType,
+    FuncExpr, FuncType, InListExpr, JoinType, LikeExpr, LogOpExpr, LogOpType, LogicalAgg,
+    LogicalEmptyRelation, LogicalFilter, LogicalJoin, LogicalLimit, LogicalProjection, LogicalScan,
+    LogicalSort, OptRelNode, OptRelNodeRef, OptRelNodeTyp, PlanNode, SortOrderExpr, SortOrderType,
 };
 
 use crate::OptdPlanContext;
 
+fn flatten_nested_logical(op: LogOpType, expr_list: ExprList) -> ExprList {
```
patrickhyw, line 23: Could you add a comment describing what this function does?

PR #109, opened 2024-03-14
feat: custom "workspaces" for generated data

patrickhyw:
**Summary**: Instead of generating data (queries, schema, etc.) directly inside the optd repo, we generate them inside a custom "workspace" directory.

**Demo**:

https://github.com/cmu-db/optd/assets/20631215/56931dce-6ac0-499f-a968-dde34f103f70

**Details**:
* The current main use case of this is to run tests locally with a "clean" workspace without needing to delete local files
* Workspaces are a CLI arg
* Relative paths are allowed and are interpreted relative to the root directory of the optd repository using `parse_pathstr()`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,8 +3,3 @@
 /.DS_Store
 /.idea
 .history
-optd-perftest/**/genned_tables
-optd-perftest/**/genned_queries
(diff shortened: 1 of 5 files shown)


Gun9niR (approved):
LGTM

PR #110, opened 2024-03-15
feat: now compatible with non-local postgres

patrickhyw:
**Summary**: Modified code so that it can work with a non-local (meaning on a different machine or in a Docker container) Postgres instance by removing all uses of `pg_ctl` and `psql`.

**Demo**:
The demo shows that the codebase has no instances of "pg_ctl" or "psql" (except for one appearance of "psql" in a comment). Despite this, `cargo run ... cardtest` still works correctly.

https://github.com/cmu-db/optd/assets/20631215/453a341a-8141-4ac4-95f1-e09b7bb690a5

**Details**:
* Removing `pg_ctl` required significant changes. Instead of having one `pgdata/` per TPC-H config's data, we now have one global `pgdata/` and one _Postgres database_ per TPC-H config's data.
* Removing `psql` involved using SQL queries to load the database instead of `psql` invocations.
* Since each TPC-H config corresponds to a Postgres database instead of a file, the string representing each TPC-H config now has additional restrictions described in `get_dbname()`.


optd-perftest/src/postgres_db.rs
```diff
@@ -1,248 +1,134 @@
...
+            )
+            .await?;
+        let pgdata_location: String = row.get("setting");
+        Ok(pgdata_location)
+    }
+
+    async fn load_benchmark_data(benchmark: &Benchmark) -> anyhow::Result<()> {
+        let dbname = benchmark.get_dbname();
+        // since we don't know whether dbname exists at this point, we have to connect to the default database
+        let default_db_client = Self::connect_to_db(DEFAULT_DBNAME).await?;
```
Gun9niR, line 72: What happens if the database does not exist?

patrickhyw: It just crashes. The error message is nice too, it says the database doesn't exist as well as the name of the database you were trying to connect to

PR #113, opened 2024-03-17
feat: integration test for cardinality performance testing

patrickhyw:
**Summary**: An integration test for cardinality performance testing.

**Demo**:

https://github.com/cmu-db/optd/assets/20631215/8eda2a04-0c23-4646-932d-8c0e22b249e4

**Details**:
* The integration test requires Postgres to be running.
* The integration test runs in its own workspace that's fully deleted at the start of each test.
* The test runs the cardtest twice to test that it works both with and without cached files in the workspace.
* Changed parts of CLI ("," delimiter for `--query-ids` instead of " ", added defaults for `--query-ids`, etc.).
* Refactored modules into `lib.rs` so that `tests/` can access all modules.
* Added `pguser` and `pgpassword` args to CLI.
* Changed COPY IN to be from STDIN instead of from a file so that we can load even to Postgres processes in a container or on another machine.


dev_scripts/kill_postgres.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+if [[ "$(hostname)" == "MacBook-Pro-20" ]] && [[ "$(whoami)" == "patrickwang" ]]; then
```
Gun9niR: Can this be removed?

dev_scripts/start_postgres.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+if [[ "$(hostname)" == "MacBook-Pro-20" ]] && [[ "$(whoami)" == "patrickwang" ]]; then
```
Gun9niR: Same

optd-perftest/src/cardtest.rs
```diff
@@ -86,3 +88,22 @@ pub trait CardtestRunnerDBHelper {
...
+    let pg_db = PostgresDb::new(workspace_dpath.clone(), pguser, pgpassword);
+    let df_db = DatafusionDb::new(workspace_dpath).await?;
+    let databases: Vec<Box<dyn CardtestRunnerDBHelper>> = vec![Box::new(pg_db), Box::new(df_db)];
+
+    let tpch_benchmark = Benchmark::Tpch(tpch_config.clone());
+    let mut cardtest_runner = CardtestRunner::new(databases).await?;
+    let qerrors = cardtest_runner
+        .eval_benchmark_qerrors_alldbs(&tpch_benchmark)
+        .await?;
+    println!("qerrors: {:?}", qerrors);
```
Gun9niR: Can we return the qerrors and let the caller decide what to do with it?

optd-perftest/tests/cardtest_integration.rs
```diff
@@ -0,0 +1,60 @@
...
+    ///   preferred way of starting it (in Docker container, with Mac app, custom build, etc.)
+    #[test]
+    fn cli_run_cardtest_twice() {
+        // perform cleanup (clear workspace)
+        let workspace_dpath = shell::parse_pathstr(WORKSPACE).unwrap();
+        shell::make_into_empty_dir(&workspace_dpath).unwrap();
+
+        // run command twice
+        for i in 1..=2 {
+            let mut cmd = create_cardtest_run_cmd(false);
```
Gun9niR: Can `cmd` be created outside of the loop?

Gun9niR (commented):
Rest LGTM

dev_scripts/kill_postgres.sh
patrickhyw: I felt like everyone would have their own preferred way of stopping/starting postgres which is why I added this

optd-perftest/src/cardtest.rs
patrickhyw: Good call

optd-perftest/tests/cardtest_integration.rs
patrickhyw: I don't think so. `output()` takes in a mutable reference so it would modify cmd.

dev_scripts/kill_postgres.sh
Gun9niR: Then perhaps not push this script? U can keep it untracked locally.

PR #115, opened 2024-03-18
Feat: simplify join cond rule and accepting all kinds of inner join conditions including non equal ones 

AveryQi115:
# Major Changes
- SimplifyJoinCondRule, just like SimplifyFilter in pr https://github.com/cmu-db/optd/pull/108
- Accepting all kinds of conditions for inner join from datafusion
    - before  this pr, optd only supports inner join on equal conditions and inner join on True/False
    - after this pr, optd now supports queries like `select x inner join y on x.a=y.a and x.a<5`. These are good testcases for push filter into join and filter push down.
    
P.S.: the core fix for merge group in this pr is based on pr https://github.com/cmu-db/optd/pull/114


jurplel (approved):
nice!

PR #118, opened 2024-03-18
Fix: fix empty relation schema maintainence across passes

AveryQi115:
# Why we need this pr?
Previously. EmptyRelation's schema is derived from its logically equivalent peer exprs. For example, the schema for `select * from empty_relation` is deducted from `select * from t1 inner join t2 on false`.  And the empty relation expression will replace the original inner join.

It works fine in the first pass as empty relation is always transformed from expressions having proper schema(except for queries like `select 64;`, it is directly converted from datafusion and has empty schema).

But it will break adaptive design as the next time query is executed, there's only empty relation left in the optimization tree and there's no way to deduct schema properties from pure empty relation.

# Design
we serialize schema as emptyRelation's data value and it is stored in its relnode, once we need the schema we fetch it from empty relation's data.

# Major Changes
- schema prop maintains the type for ColumnRef typ and List typ
- adds convertion from Value to ConstantType to convert data to schema field
- adds serialized data to EmptyRelation data so schema is keeped
- change EmptyRelation initialization method
- change columnExpr derivation rules for emptyRelation

### DIFF
--- a/optd-datafusion-repr/src/properties/schema.rs
+++ b/optd-datafusion-repr/src/properties/schema.rs
@@ -3,8 +3,8 @@ use std::sync::Arc;
 
 use optd_core::property::PropertyBuilder;
 
-use crate::plan_nodes::{ConstantType, EmptyRelationData, OptRelNodeTyp};
 use super::DEFAULT_NAME;
+use crate::plan_nodes::{ConstantType, EmptyRelationData, OptRelNodeTyp};
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct Field {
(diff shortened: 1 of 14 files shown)


optd-datafusion-repr/src/properties/schema.rs
```diff
@@ -61,18 +62,28 @@ impl PropertyBuilder<OptRelNodeTyp> for SchemaPropertyBuilder {
...
-                    fields: vec![
-                        Field {
-                            name: "unnamed".to_string(),
-                            typ: ConstantType::Any,
-                            nullable: true
-                        };
-                        children.len()
-                    ],
+                    fields: vec![Field {
+                        name: "unnamed".to_string(),
```
Gun9niR: Perhaps make `unnamed` a constant?

Gun9niR (approved):
Rest LGTM

PR #119, opened 2024-03-20
Fix: Fix for physical when budget is reached

AveryQi115:
Continue on executing physical rules when budget limit is reached.

### DIFF
--- a/optd-sqlplannertest/tests/tpch.planner.sql
+++ b/optd-sqlplannertest/tests/tpch.planner.sql
@@ -364,72 +364,69 @@ PhysicalLimit { skip: 0, fetch: 100 }
     │   └── SortOrder { order: Asc }
     │       └── #3
     └── PhysicalProjection { exprs: [ #5, #2, #8, #0, #1, #3, #4, #6 ] }
-        └── PhysicalProjection { exprs: [ #0, #1, #2, #3, #4, #5, #6, #7, #8, #11, #12 ] }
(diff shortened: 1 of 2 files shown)


skyzh (approved):
🤪 thanks for the fix!

optd-sqlplannertest/tests/tpch.planner.sql
```diff
@@ -364,72 +364,69 @@ PhysicalLimit { skip: 0, fetch: 100 }
...
-                                                                                                        │   └── PhysicalScan { table: nation }
-                                                                                                        └── PhysicalProjection { exprs: [ #0 ] }
-                                                                                                            └── PhysicalFilter
-                                                                                                                ├── cond:Eq
-                                                                                                                │   ├── #1
-                                                                                                                │   └── "AFRICA"
-                                                                                                                └── PhysicalProjection { exprs: [ #0, #1 ] }
-                                                                                                                    └── PhysicalScan { table: region }
+        └── PhysicalNestedLoopJoin
```
skyzh, line 367: NLJ not optimized into hash join?

skyzh: or something wrong with the cost model that it assumes NLJ costs less than hash join

AveryQi115: Oh it's due to our physical conversion rule of hashjoin. It only takes one pair of equal keys so I assume nlj is more efficient than multiple layers of projection and hashjoin. See the previous solution, it contains a lot more projection and hash join. I'll merge this pr first and modify current hash join later.

PR #122, opened 2024-03-21
feat: printing q-errors nicely + explains

patrickhyw:
**Summary**: Now printing aggregate q-errors, per-query q-errors, and EXPLAIN results to facilitate experimentation.

**Demo**:
Screenshot 2024-03-21 at 13 34 13
Screenshot 2024-03-21 at 13 28 24

**Details**:
* Automatically prints the `EXPLAIN` results of all queries from all DBs to info. Use `RUST_LOG=info` to see them. Use `--query-ids X` to just investigate one specific query.
* Now handles edge case where file indicating that a db was created exists but the db doesn't actually exist.
* Gracefully handles the case of infinite q-errors.
* Deleted dev_scripts. Moved to [gungnir-experiments](https://github.com/wangpatrick57/gungnir-experiments).
* Uses `prettytable-rs` for printing.


Gun9niR (commented):
There must be some crate that we can use to print table... no need to manually write `| max = ... | min = ... |

Gun9niR (approved):

PR #124, opened 2024-03-21
feat: caching true cardinalities, 22x speedup for TPC-H SF1

patrickhyw:
**Summary**: Instead of executing queries every time we run cardtest (takes multiple minutes even for scale factor 1.0), we cache the true cardinalities from the first execution.

**Demo**:
22x speedup (29m55s -> 1m21s) on TPC-H scale factor 1

Screenshot 2024-03-21 at 21 06 10
Screenshot 2024-03-21 at 21 05 32

JSON cache file
Screenshot 2024-03-21 at 19 00 09

**Details**:
* We now call `VACUUM FULL ANALYZE` in Postgres right after loading data to create stats, which we weren't doing before.
* Because Postgres is faster and less buggy than optd+DataFusion, we only get truecards from Postgres.
* Fixed bug where `dbname` and `tables/` were parameterized by the seed, which would make us miss caching opportunities.

### DIFF
--- a/optd-perftest/src/postgres_dbms.rs
+++ b/optd-perftest/src/postgres_dbms.rs
@@ -160,7 +160,7 @@ impl PostgresDBMS {
 
         // create stats
         // you need to do VACUUM FULL ANALYZE and not just ANALYZE to make sure the stats are created in a deterministic way
-        // this is standard practice for ostgres benchmarking
+        // this is standard practice for postgres benchmarking
         client.query("VACUUM FULL ANALYZE", &[]).await?;
 
         Ok(())
--- a/optd-perftest/src/postgres_dbms.rs
+++ b/optd-perftest/src/postgres_dbms.rs
@@ -37,7 +37,11 @@ pub struct PostgresDBMS {
 ///   - Stop and start functions should be separate
 ///   - Setup should be done in build() unless it requires more information (like benchmark)
 impl PostgresDBMS {
-    pub fn build<P: AsRef<Path>>(workspace_dpath: P, pguser: &str, pgpassword: &str) -> anyhow::Result<Self> {
+    pub fn build<P: AsRef<Path>>(
+        workspace_dpath: P,
+        pguser: &str,
+        pgpassword: &str,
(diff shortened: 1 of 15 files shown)


optd-perftest/src/postgres_dbms.rs
```diff
@@ -145,6 +158,11 @@ impl PostgresDb {
             Self::copy_from_stdin(client, tbl_fpath).await?;
         }
 
+        // create stats
+        // you need to do VACUUM FULL ANALYZE and not just ANALYZE to make sure the stats are created in a deterministic way
+        // this is standard practice for ostgres benchmarking
```
Gun9niR:
```suggestion
        // this is standard practice for postgres benchmarking
```

optd-perftest/src/truecard_cache.rs
```diff
@@ -1,13 +1,81 @@
...
+    dbms_cache: HashMap<String, HashMap<String, usize>>,
 }
 
-// TODO(phw2): write function that takes in a Benchmark and creates a new entry in the cache
-impl TruecardCache {
-    
-}
\ No newline at end of file
+impl DBMSTruecardCache {
+    fn get_ser_fpath<P: AsRef<Path>>(workspace_dpath: P, dbms_name: &str) -> PathBuf {
```
Gun9niR: What does `ser` mean?

optd-perftest/src/truecard_cache.rs
```diff
@@ -1,13 +1,81 @@
...
+            }
+        };
+        db_cache.insert(String::from(sql), truecard);
+    }
+
+    pub fn get_truecard(&self, dbname: &str, sql: &str) -> Option<usize> {
+        match self.dbms_cache.get(dbname) {
+            Some(db_cache) => db_cache.get(sql).copied(),
+            None => None,
+        }
```
Gun9niR:
```suggestion
        self.dbms_cache
            .get(dbname)
            .map(|db_cache| db_cache.get(sql).copied())
            .flatten()
```

Gun9niR (approved):
Rest LGTM

PR #125, opened 2024-03-22
feat: Extend scan node to support additional fields from DataFusion

jurplel:
Moves limits, filtering, and projection into the scan node representation. In the process, I moved the scan node to use macros, which required modifying the macros to support an optional data field.

### DIFF
--- a/optd-datafusion-bridge/src/from_optd.rs
+++ b/optd-datafusion-bridge/src/from_optd.rs
@@ -73,10 +73,38 @@ impl OptdPlanContext<'_> {
         &mut self,
         node: PhysicalScan,
     ) -> Result<Arc<dyn ExecutionPlan + 'static>> {
-        let table = node.into_rel_node().data.as_ref().unwrap().as_str();
+        let table = node.clone().into_rel_node().data.as_ref().unwrap().as_str();
         let source = self.tables.get(table.as_ref()).unwrap();
         let provider = source_as_provider(source)?;
-        let plan = provider.scan(self.session_state, None, &[], None).await?;
+
+        let proj_expr_list = node
+            .projections()
+            .to_vec()
+            .into_iter()
+            .map(|expr| {
+                ConstantExpr::from_rel_node(expr.into_rel_node())
+                    .unwrap()
+                    .value()
+                    .as_u64() as usize
+            })
+            .collect();
+
+        assert!(node.fetch().typ() == OptRelNodeTyp::Constant(ConstantType::UInt64));
+        let fetch = ConstantExpr::from_rel_node(node.fetch().into_rel_node())
+            .unwrap()
+            .value()
+            .as_u64();
+        let fetch_opt: Option<usize> = if fetch == u64::MAX {
+            None
+        } else {
+            Some(fetch.try_into().unwrap())
+        };
+
+        // TODO: Filter expr can't be converted easily to this Expr type
+
+        let plan = provider
+            .scan(self.session_state, Some(&proj_expr_list), &[], fetch_opt)
+            .await?;
         Ok(plan)
     }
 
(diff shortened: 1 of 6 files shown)


optd-datafusion-repr/src/plan_nodes/macros.rs
```diff
@@ -44,16 +45,24 @@ macro_rules! define_plan_node {
             pub fn new(
                 $($child_name : $child_meta_typ,)*
                 $($attr_name : $attr_meta_typ),*
+                $(, $data_name : Value)?
                 $(, $inner_name : $inner_typ)?
             ) -> $struct_name {
+                #[allow(unused_mut, unused)]
```
jurplel, line 51: If anyone can find a way to conditionally insert code in an if/else pattern, or otherwise avoid allowing these warnings, let me know!

skyzh:
I'm leaning towards the current simplified plan node design that scan only does scans. The PhysicalScan node can have as much information as you want and there can be a rule to do that. Otherwise I could imagine the push down rules would be much more complex as it will need to handle the merge of seqscan and other plan nodes instead of simply doing a push down.

jurplel:
> I'm leaning towards the current simplified plan node design that scan only does scans. The PhysicalScan node can have as much information as you want and there can be a rule to do that. Otherwise I could imagine the push down rules would be much more complex as it will need to handle the merge of seqscan and other plan nodes instead of simply doing a push down.

What originally prompted me to make this change was that filter pushdown and projection pushdown need to have a way to push the filters and projections into the scan node itself, so that we don't pass intermediate data around.

Are you suggesting that the logical->physical transformation rule detects projections and filters above the scan node, merging them in one move? I am not sure how to do that, and it seems to me like it might be more complex.

The most complex parts of filter pushdown that I have encountered so far involve separating out conjunctions/pushing partial predicates. It would be fairly easy for me to push into a scan node, even with an existing predicate.

optd-datafusion-repr/src/plan_nodes/macros.rs
```diff
@@ -5,6 +5,7 @@ macro_rules! define_plan_node {
         [ $({ $child_id:literal, $child_name:ident : $child_meta_typ:ty }),* ] ,
         [ $({ $attr_id:literal, $attr_name:ident : $attr_meta_typ:ty }),* ]
         $(, { $inner_name:ident : $inner_typ:ty })?
+        $(, $data_name:ident)?
```
Gun9niR, line 8: Need to take this into account in `dispatch_explain`. 

Another problem I can think of is when data is `Value::Serialized`, because it makes no sense to print serialized data. We can prob add another `$(, $data_type:path)?` to be matched and check `data` against `Value::Serialized` explicitly.

optd-datafusion-bridge/src/from_optd.rs
```diff
@@ -73,7 +73,10 @@ impl OptdPlanContext<'_> {
         &mut self,
         node: PhysicalScan,
     ) -> Result<Arc<dyn ExecutionPlan + 'static>> {
-        let source = self.tables.get(node.table().as_ref()).unwrap();
+        let table = node.into_rel_node().data.as_ref().unwrap().as_str();
+        dbg!(table.clone());
+        dbg!(self.tables.clone().keys());
+        let source = self.tables.get(table.as_ref()).unwrap();
         let provider = source_as_provider(source)?;
         let plan = provider.scan(self.session_state, None, &[], None).await?;
```
jurplel: this was not sufficiently modified

optd-datafusion-repr/src/plan_nodes/macros.rs
Gun9niR: I can try to handle that

jurplel: #129 for the paper trail

optd-datafusion-repr/src/plan_nodes/macros.rs
jurplel: Direct to #129

jurplel:
will pick up again another day

PR #128, opened 2024-03-22
feat: new TruecardGetter interface

patrickhyw:
**Summary**: Made a new `TruecardGetter` interface that allows individual DBMS cardtest helpers to _not_ implement query execution.

**Details**:
* This allows us to decouple stats creation from table loading in DataFusion, causing TPC-H SF1's cardtest to run 1.7x faster.
* More importantly, this sets up a future PR where we _cache_ the statistics object in DataFusion, which would unlock a ~10x speedup for TPC-H SF1's cardtest.
* Q6 also now works with these changes so I added it to the default args.


optd-perftest/src/cardtest.rs
```diff
@@ -62,54 +70,46 @@ impl CardtestRunner {
...
+///   for the same conceptual query (see how qgen in tpch-kit takes in DBMS as an input).
 /// When more performance tests are implemented, you would probably want to extract
 ///   get_name() into a generic "DBMS" trait.
 #[async_trait]
 pub trait CardtestRunnerDBMSHelper {
     // get_name() has &self so that we're able to do Box<dyn CardtestRunnerDBMSHelper>
     fn get_name(&self) -> &str;
 
-    // The order of queries has to be the same between these two functions.
+    // The order of queries has to be the same between all databases
```
Gun9niR: `Benchmark` hides the notion of "order of queries" right

optd-perftest/src/postgres_dbms.rs
```diff
@@ -318,3 +283,48 @@ impl PostgresDBMS {
...
+        match benchmark {
+            Benchmark::Test => unimplemented!(),
+            Benchmark::Tpch(tpch_config) => self.eval_tpch_estcards(&client, tpch_config).await,
+        }
+    }
+}
+
+#[async_trait]
+impl TruecardGetter for PostgresDBMS {
+    async fn get_benchmark_truecards(&mut self, benchmark: &Benchmark) -> anyhow::Result<Vec<usize>> {
```
Gun9niR: Should this function take a cache as a parameter? Because all `TruecardGetter` instances should share the same cache?

optd-perftest/src/cardtest.rs
patrickhyw: Oh I meant the order of queries in the returned vector. I'll clarify that in the comment

optd-perftest/src/postgres_dbms.rs
patrickhyw: I was thinking to abstract the cache away from the user so they don't have to worry about it. Every time you call this function, it reads from the file, updates the cache, and writes back to the file, so if you call it multiple times it is shared in that way.

Gun9niR: That makes sense

optd-perftest/src/datafusion_dbms.rs
```diff
@@ -230,6 +208,18 @@ impl DatafusionDBMS {
         for ddl in ddls {
             Self::execute(&self.ctx, ddl).await?;
         }
+        Ok(())
+    }
+
+    async fn _load_tpch_data_no_stats(&mut self, tpch_config: &TpchConfig) -> anyhow::Result<()> {
```
Gun9niR, line 214:
```suggestion
    #[allow(dead_code)]
    async fn load_tpch_data_no_stats(&mut self, tpch_config: &TpchConfig) -> anyhow::Result<()> {
```

optd-perftest/src/datafusion_dbms.rs
```diff
@@ -315,7 +310,16 @@ impl DatafusionDBMS {
             );
             log::debug!("statistics generated for table: {}", tbl_name);
         }
-        Ok(base_table_stats)
+
+        self.clear_state(Some(base_table_stats)).await?;
+
+        // Create the tables. This must be done after creating stats.
+        self.create_tpch_tables(&tpch_kit).await?;
```
Gun9niR, line 317: Is it possible to move the side effects out?

Gun9niR (approved):
Rest LGTM

PR #129, opened 2024-03-22
feat: add data field in define_plan_node macro

Gun9niR:
Add a `data_name` field to `define_plan_node` macro. Plan nodes with data field must implement a method `explain_data`. 

**An example**

```rust
#[derive(Clone, Debug, Serialize, Deserialize)]
struct ComplexData {
    a: i32,
    b: String,
}

#[derive(Clone, Debug)]
struct PhysicalComplexDummy(PlanNode);

impl PhysicalComplexDummy {
    pub fn explain_data(data: &Value) -> Vec<(&'static str, Pretty<'static>)> {
        if let Value::Serialized(serialized_data) = data {
            let data: ComplexData = bincode::deserialize(serialized_data).unwrap();
            vec![
                ("a", data.a.to_string().into()),
                ("b", data.b.to_string().into()),
            ]
        } else {
            unreachable!()
        }
    }
}

define_plan_node!(
    PhysicalComplexDummy: PlanNode,
    PhysicalScan, [
        { 0, child: PlanNode }
    ], [
    ],
    complex_data
);
```

I could just print data with `format!("{:?}, data)`, but the downside of this is that for a struct 

```rust
struct Foo {
    a: i32,
    b: i32,
}
```

The debug output would be `Foo { a: 1, b: 2}`. Here, the struct name is implementation detail that should be hidden from the user, and the bracket surrounding the fields makes these fields seem related in some way, when they are actually not. So just letting user implement `explain_data` gives maximum flexibility.


optd-datafusion-repr/src/plan_nodes/macros.rs
```diff
@@ -1,10 +1,29 @@
...
+/// impl ExplainData for PhysicalDummy {
+///     fn explain_data(data: &Value) -> Vec<(&'static str, Pretty<'static>)> {
+///         if let Value::Int32(i) = data {
+///             vec![("primitive_data", i.to_string().into())]
+///         } else {
+///             unreachable!()
+///         }
+///     }
+/// }
+/// ```
```
jurplel, line 18: Instead of requiring this trait implementation would it be possible to instead allow the macro to take in a type for the data argument? Then, the explain function could be generated to automatically cast it from value to the correct type, without requiring any trait implementation.

jurplel (approved):
Looks good overall, one possible design suggestion from me

optd-datafusion-repr/src/plan_nodes/macros.rs
Gun9niR: How would you know which `Value`'s variant it is then? Besides, if data is a struct we don't know how it's serialized, and how the fields should be displayed.

jurplel: You're right. I was thinking of the standard integer/string types. It seems a shame to me to have a whole trait for this, but it's a reasonable solution.

PR #130, opened 2024-03-22
feat: new TruecardGetter interface

patrickhyw:
**Summary**: Made a new `TruecardGetter` interface that allows individual DBMS cardtest helpers to _not_ implement query execution.

**Details**:
* This allows us to decouple stats creation from table loading in DataFusion, causing TPC-H SF1's cardtest to run 1.7x faster.
* More importantly, this sets up a future PR where we _cache_ the statistics object in DataFusion, which would unlock a ~10x speedup for TPC-H SF1's cardtest.
* Q6 also now works with these changes so I added it to the default args.


optd-perftest/src/postgres_dbms.rs
```diff
@@ -158,11 +164,20 @@ impl PostgresDBMS {
             Self::copy_from_stdin(client, tbl_fpath).await?;
         }
 
+        // load the constraints
+        let tpch_constraints_fpath = get_tpch_constraints_fpath().unwrap();
+        println!("tpch_constraints_fpath={:?}", tpch_constraints_fpath);
+        let sql = fs::read_to_string(tpch_constraints_fpath.to_str().unwrap())?;
+        client.batch_execute(&sql).await?;
```
Gun9niR: Why do we need this?

Gun9niR (approved):

PR #131, opened 2024-03-22
feat: load indexes for TPC-H, 13x speedup on SF 0.1

patrickhyw:
**Summary**: Now loading constraints + indexes (though constraints are temporarily broken) when loading data for TPC-H.

**Demo**:
Without constraints, TPC-H SF0.1 takes ~13s. With constraints, it takes ~1s.
Screenshot 2024-03-22 at 17 30 15
Screenshot 2024-03-22 at 17 31 19

**Details**:
* Constraints file is from dbgym (is a private repo so I can't link it).
* Note that constraints are currently broken so I commented it. Not adding constraints doesn't cause noticeable issues so I'm punting this to later.
* Made `get_optd_root()` check that it's actually in the optd repo (and not the tpch-kit repo).
* Added option to run commands in a directory instead of manually changing directory beforehand (which might cause us to CD into tpch-kit and mess up `get_optd_root()`).


Gun9niR:
The dbgym link is broken?

optd-perftest/src/tpch_constraints.sql
Gun9niR, line 1: Isn't this more like a create index file instead of constraints file?

patrickhyw: Fixed

patrickhyw: I realized dbgym is a private repo. I'll remove the link

Gun9niR (approved):
lgtm

PR #132, opened 2024-03-23
feat: caching optd stats, 12x speedup on TPC-H SF1

patrickhyw:
**Summary**: Now caching the stat objects used by `OptCostModel`, meaning we don't need to load data into DataFusion after doing it the first time.

**Demo**:
12x speedup on TPC-H SF1 compared to not caching stats.

Caching everything _except_ optd stats takes 45.6s total.
Screenshot 2024-03-23 at 16 59 04

Caching everything, _including_ optd stats, takes 3.9s total.
Screenshot 2024-03-23 at 16 57 45

**Details**:
* This caching is **disabled by default** to avoid accidentally using stale stats. I added a CLI arg to enable it.
* The main challenge of this PR was making `PerTableStats` a serializable object for `serde`.
* The serializability refactor will also help down the line when we want to **put statistics in the catalog**, since that is fundamentally a serialization problem too. Having `Box<dyn ...>` would make putting stats in the catalog more difficult.
* This required a significant refactor of how the `MostCommonValues` and `Distribution` traits are handled in `OptCostModel`. Instead of having `Box<dyn ...>` values in `PerColumnStats` which store any object that implements these traits, I made `PerColumnStats` a templated object.
* The one downside of this refactor is that we can no longer have a database which uses _different_ data structures for `Distribution` (like a t-digest for one column, a histogram for another, etc.). I didn't see this as a big enough reason to not do the refactor because it seems like a rare thing to do. Additionally, if we really needed to do this, we could just make an enum that had both types.

### DIFF
--- a/optd-datafusion-repr/src/cost/base_cost.rs
+++ b/optd-datafusion-repr/src/cost/base_cost.rs
@@ -36,13 +36,21 @@ fn compute_plan_node_cost<T: RelNodeTyp, C: CostModel<T>>(
     cost
 }
 
-pub type BaseTableStats = HashMap<String, PerTableStats>;
+pub type BaseTableStats<M, D> = HashMap<String, PerTableStats<M, D>>;
 
-pub struct OptCostModel {
-    per_table_stats_map: BaseTableStats,
+// The "standard" concrete types that optd currently uses
+// All of optd (except unit tests) must use the same types
+pub type StandardMostCommonValues = MockMostCommonValues;
+pub type StandardDistribution = TDigest;
+pub type StandardBaseTableStats = BaseTableStats<StandardMostCommonValues, StandardDistribution>;
+pub type StandardPerTableStats = PerTableStats<StandardMostCommonValues, StandardDistribution>;
+pub type StandardPerColumnStats = PerColumnStats<StandardMostCommonValues, StandardDistribution>;
+
+pub struct OptCostModel<M: MostCommonValues, D: Distribution> {
+    per_table_stats_map: BaseTableStats<M, D>,
 }
 
-struct MockMostCommonValues {
+pub struct MockMostCommonValues {
     mcvs: HashMap<Value, f64>,
 }
 
@@ -76,12 +84,12 @@ impl MostCommonValues for MockMostCommonValues {
     }
 }
 
-pub struct PerTableStats {
+pub struct PerTableStats<M: MostCommonValues, D: Distribution> {
     row_cnt: usize,
-    per_column_stats_vec: Vec<Option<PerColumnStats>>,
+    per_column_stats_vec: Vec<Option<PerColumnStats<M, D>>>,
 }
 
-impl PerTableStats {
+impl StandardPerTableStats {
     pub fn from_record_batches<I: IntoIterator<Item = Result<RecordBatch, ArrowError>>>(
         batch_iter: RecordBatchIterator<I>,
     ) -> anyhow::Result<Self> {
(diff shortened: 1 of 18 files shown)


optd-datafusion-repr/src/cost/base_cost.rs
```diff
@@ -36,13 +37,22 @@ fn compute_plan_node_cost<T: RelNodeTyp, C: CostModel<T>>(
...
 
-pub struct OptCostModel {
-    per_table_stats_map: BaseTableStats,
+// The "standard" concrete types that optd currently uses
+// All of optd (except unit tests) must use the same types
+pub type StandardMostCommonValues = MockMostCommonValues;
+pub type StandardDistribution = TDigest;
+pub type StandardBaseTableStats = BaseTableStats<StandardMostCommonValues, StandardDistribution>;
+pub type StandardPerTableStats = PerTableStats<StandardMostCommonValues, StandardDistribution>;
+pub type StandardPerColumnStats = PerColumnStats<StandardMostCommonValues, StandardDistribution>;
```
Gun9niR: Maybe use `Datafusion` instead of `Standard`? It feels more informative.

optd-perftest/src/datafusion_dbms.rs
```diff
@@ -275,53 +288,73 @@ impl DatafusionDBMS {
...
+        tpch_config: &TpchConfig,
+        benchmark_fname: String,
+    ) -> anyhow::Result<StandardBaseTableStats> {
+        let stats_cache_fpath = self
+            .workspace_dpath
+            .join("datafusion_stats_caches")
+            .join(format!("{}.json", benchmark_fname));
+        if self.use_stats_cache && stats_cache_fpath.exists() {
+            let file = File::open(&stats_cache_fpath)?;
+            Ok(serde_json::from_reader(file)?)
```
Gun9niR: This logic applies to every benchmark right? So can it be moved into `get_benchmark_stats`?

optd-perftest/src/datafusion_dbms.rs
```diff
@@ -275,53 +288,73 @@ impl DatafusionDBMS {
...
+            }
 
-        // Generate the tables
-        let tpch_kit = TpchKit::build(&self.workspace_dpath)?;
-        tpch_kit.gen_tables(tpch_config)?;
+            if self.use_stats_cache {
+                fs::create_dir_all(stats_cache_fpath.parent().unwrap())?;
+                let file = File::create(&stats_cache_fpath)?;
+                serde_json::to_writer(file, &base_table_stats)?;
+            }
```
Gun9niR: ditto

optd-perftest/tests/cardtest_integration.rs
```diff
@@ -44,6 +44,7 @@ mod tests {
             // make sure scale factor is low so the test runs fast
             "--scale-factor",
             "0.01",
+            "--use-optd-stats-cache",
```
Gun9niR: Is there a reason to not use the cache?

optd-perftest/src/benchmark.rs
```diff
@@ -37,6 +37,17 @@ impl Benchmark {
         dbname.to_lowercase()
     }
 
+    /// Use this when you need a file name. The rules for file names are different from the rules
+    ///   for database names, so this is a different function
+    pub fn get_fname(&self) -> String {
```
Gun9niR, line 42: More generally, it can be unique name

Gun9niR (commented):
Rest LGTM

optd-perftest/src/datafusion_dbms.rs
patrickhyw: Good catch yeah, I moved it there.

optd-perftest/tests/cardtest_integration.rs
patrickhyw: If the stats are updated (e.g. algorithmic improvement to T-Digest) you want to regenerate them instead of using the cache. To avoid accidentally using stale stats, I make it not read from the cache by default.

optd-perftest/src/benchmark.rs
patrickhyw: Hm, wdym?

Gun9niR: I mean it doesn't have to be a file name. The function's name is a bit confusing, like: file for what? But actually it's just a unique name that can be used as a file name

Gun9niR (approved):

PR #134, opened 2024-03-24
feat: add an arithmetic encoder to transform strings into f64

AlSchlo:
This PR implements an arithmetic encoder for alphanumeric strings, enabling the strings to utilize t-digests (see [Wikipedia](https://en.wikipedia.org/wiki/Arithmetic_coding)). 

Characters outside the specified alpha-numerical range are treated as identical and considered 'heaviest'.


Gun9niR (approved):
Very cool!

PR #136, opened 2024-03-24
refactor: add data type to `define_plan_node` macro

Gun9niR:
Add data type to `define_plan_node`. The benefits include:

- Pass actual data type in `new` instead of converting it to `Value` outside first to reduce boilerplate code.
- Aggregate the serialization and deserialization to one place, to ensure serialization and deserialization use the same format
- Data type is clear from the macro invocation

### DIFF
--- a/optd-datafusion-repr/src/plan_nodes.rs
+++ b/optd-datafusion-repr/src/plan_nodes.rs
@@ -208,6 +208,13 @@ pub trait OptRelNode: 'static + Clone {
     }
 }
 
+/// Plan nodes that are defined through `define_plan_node` macro with data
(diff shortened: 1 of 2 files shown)


jurplel (approved):

yliang412:
@Gun9niR Forgot to mention it here:
- Is it possible to have a default `explain_data `implementation for `T: Display`?
- Maybe also `T: From<Value> + Into<Value>`?

Gun9niR:
> @Gun9niR Forgot to mention it here:
> 
> * Is it possible to have a default `explain_data `implementation for `T: Display`?
> * Maybe also `T: From<Value> + Into<Value>`?


Good idea!

PR #137, opened 2024-03-25
feat: move data and value conversion from `ExplainData` to `From`

Gun9niR:
Move `value_to_data` and `data_to_value` to `From<&Value>` and `From<DataType>`


jurplel (approved):
nice!

yliang412 (approved):
LGTM

PR #139, opened 2024-03-26
Feat: two stage heuristic

AveryQi115:
This pr adds infer properties for heuristic optimizer in optd core.

It splits the two passes for heuristic and cascades. Now heuristic rules can be used on the first heuristic optimizer as registering them as default_heuristic_rule and these rules are able to traverse the whole plan tree as there are no placeholder in heuristic stage.

The properties are not shared among heuristic pass and cascades pass though. It is recalculated in cascades.

The pr is for enabling more complicated heuristic rules and one can register their rules as either heuristic or cascade to test its usability and performance.

The logic for the rule creation is the same as previous heuristic rule wrapper:
heuristic rule either return 0 or 1 node. Returning 1 node means that the rule is successfully applied. Returning 0 node means it fails some constraints for the rule and heuristic optimizer will use the original node.

### DIFF
--- a/optd-core/src/heuristics/optimizer.rs
+++ b/optd-core/src/heuristics/optimizer.rs
@@ -1,9 +1,12 @@
 use std::{collections::HashMap, sync::Arc};
 
 use anyhow::Result;
+use itertools::Itertools;
+use std::any::Any;
 
 use crate::{
     optimizer::Optimizer,
+    property::PropertyBuilderAny,
     rel_node::{RelNode, RelNodeRef, RelNodeTyp},
     rules::{Rule, RuleMatcher},
 };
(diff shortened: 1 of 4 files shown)


jurplel:
All looks good to me at a glance, great job making this PR so quickly.

jurplel (approved):

Sweetsuro (approved):
LGTM, thanks Avery!

optd-core/src/heuristics/optimizer.rs
```diff
@@ -141,20 +154,60 @@ impl<T: RelNodeTyp> HeuristicsOptimizer<T> {
...
                     .into(),
                 )?;
+                self.infer_properties(root_rel.clone());
+                self.properties.insert(
+                    node.clone(),
+                    self.properties.get(&root_rel.clone()).unwrap().clone(),
+                );
                 Ok(node)
             }
             ApplyOrder::TopDown => {
```
jurplel, line 164: Something I ran into while testing filter pushdown—to apply rules top-down that use properties, it seems that properties need to be inferred before the rule is applied.

It also needs to be inferred again after the rule is applied, since the children change.

Adding `self.infer_properties(root_rel.clone());` to the first line of this block, before `apply_rules`, fixed the issue for me.

AveryQi115: Thanks for catching that!

PR #140, opened 2024-03-26
feat: Filter Pushdown Rule

jurplel:
This PR brings a filter pushdown heuristic rule, built on @AveryQi115's hybrid scheme.

### Filter Pushdown Rule
- This series of rules matches on any filter, and pushes any part of the filter down below a node if possible.
- They are registered in the HeuristicsRuleWrapper currently, but they work as cost-based rules as well.

### Helper Functions
- `LogOpExpr::new_flattened_nested_logical` creates a new `LogOpExpr` from an `ExprList`, and it flattens any nested `LogOpExpr`s of the same `LogOpType`.
- `Expr::rewrite_column_refs` recursively rewrites any `ColumnExpr` in an expression tree, using a provided `rewrite_fn`.
- `LogicalJoin::map_through_join` takes in left/right schema sizes, and maps an index to be as it would if it were pushed down to the left or right side of a join.
- `LogicalProjection::compute_column_mapping` creates a `ColumnMapping` object from a `LogicalProjection`. 
    - The `ColumnMapping` object has a few methods, but most importantly it has `rewrite_condition`, which given an expr, will rewrite the expression with the projection's mapping.

### Testing Utilities
- `new_test_optimizer` creates a new heuristic optimizer, which applies a given rule. It uses a `TpchCatalog`.
- `TpchCatalog` is a catalog implementing a couple of tables from the TPC-H schema. It can be extended to have more as needed.
- `DummyCostModel` implements a cost model, only giving zero cost. It is used for constructing a cascades optimizer without a real cost model, and isn't used in this PR. 
- This pull request, using these test optimizer components, pioneers a new testing scheme, based on running a constructed query plan through an optimizer, rather than text-based SQL planner tests, which may be flaky. They also test rules in isolation.


yliang412:
I was looking at Umbra's plan For Q7. Would it be helpful to push down hints for the `(n1.n_name = 'FRANCE' AND n2.n_name = 'GERMANY')
                OR (n1.n_name = 'GERMANY' AND n2.n_name = 'FRANCE')` predicate? You still need to keep the join, but you could push down `FRANCE` or `GERMANY` to both sides and reduce the number of things to look at.



TPC-H Q7:
```sql
SELECT
    supp_nation,
    cust_nation,
    l_year,
    SUM(volume) AS revenue
FROM
    (
        SELECT
            n1.n_name AS supp_nation,
            n2.n_name AS cust_nation,
            EXTRACT(YEAR FROM l_shipdate) AS l_year,
            l_extendedprice * (1 - l_discount) AS volume
        FROM
            supplier,
            lineitem,
            orders,
            customer,
            nation n1,
            nation n2
        WHERE
            s_suppkey = l_suppkey
            AND o_orderkey = l_orderkey
            AND c_custkey = o_custkey
            AND s_nationkey = n1.n_nationkey
            AND c_nationkey = n2.n_nationkey
            AND (
                (n1.n_name = 'FRANCE' AND n2.n_name = 'GERMANY')
                OR (n1.n_name = 'GERMANY' AND n2.n_name = 'FRANCE')
            )
            AND l_shipdate BETWEEN DATE '1995-01-01' AND DATE '1996-12-31'
    ) AS shipping
GROUP BY
    supp_nation,
    cust_nation,
    l_year
ORDER BY
    supp_nation,
    cust_nation,
    l_year;
```

jurplel:
> I was looking at Umbra's plan For Q7. Would it be helpful to push down hints for the `(n1.n_name = 'FRANCE' AND n2.n_name = 'GERMANY') OR (n1.n_name = 'GERMANY' AND n2.n_name = 'FRANCE')` predicate? You still need to keep the join, but you could push down `FRANCE` or `GERMANY` to both sides and reduce the number of things to look at.
> ...

This is a great example, but if you look at what Umbra is doing, this is actually a separate step from pushdown:
**Unoptimized:**
image
**Expression Simplification:**
image

If the filter were simplified to a conjunction first, then the filter pushdown implementation would be able to operate on the clauses independently. Dealing with an `in` expression with two columns is a separate issue, but I'm not sure optd would simplify the expression like that.

yliang412:
> > I was looking at Umbra's plan For Q7. Would it be helpful to push down hints for the `(n1.n_name = 'FRANCE' AND n2.n_name = 'GERMANY') OR (n1.n_name = 'GERMANY' AND n2.n_name = 'FRANCE')` predicate? You still need to keep the join, but you could push down `FRANCE` or `GERMANY` to both sides and reduce the number of things to look at.
> > ...
> 
> This is a great example, but if you look at what Umbra is doing, this is actually a separate step from pushdown: **Unoptimized:** image **Expression Simplification:** image
> 
> If the filter were simplified to a conjunction first, then the filter pushdown implementation would be able to operate on the clauses independently. Dealing with an `in` expression with two columns is a separate issue, but I'm not sure optd would simplify the expression like that.

Yep, I agree that this seems to be relevant to expression optimizations.

optd-datafusion-repr/src/lib.rs
```diff
@@ -29,11 +29,18 @@ use rules::{
 
 pub use optd_core::rel_node::Value;
 
+use crate::rules::{
```
Sweetsuro: I guess we'll need more transpose rules after support for left semi join + left anti join are added, add to some issue (?) You should probably also put these rules in with the other rules instead of in their own spot

optd-datafusion-repr/src/plan_nodes/join.rs
```diff
@@ -62,3 +62,20 @@ define_plan_node!(
         { 3, right_keys: ExprList }
     ], { join_type: JoinType }
 );
+
+impl LogicalJoin {
+    /// Takes in left/right schema sizes, and maps an index to be as if it were
```
Sweetsuro, line 67: What is an index?

optd-datafusion-repr/src/plan_nodes/join.rs
```diff
@@ -62,3 +62,20 @@ define_plan_node!(
...
+    /// pushed down to the left or right side of a join accordingly.
+    pub fn map_through_join(
+        index: usize,
```
Sweetsuro: I guess change variable name + comment to specify column index?

optd-datafusion-repr/src/plan_nodes/projection.rs
```diff
@@ -26,3 +26,57 @@ define_plan_node!(
...
+    pub fn projection_col_refers_to(&self, col: usize) -> usize {
+        self.forward[col]
+    }
+
+    pub fn _original_col_maps_to(&self, col: usize) -> Option<usize> {
+        self._backward[col]
+    }
+
+    /// Recursively rewrites all ColumnRefs in an Expr to what the projection
+    /// node is rewriting. E.g. if Projection is A -> B, B will be rewritten as A
```
Sweetsuro: Kind of unclear, maybe putting an example would help?

optd-datafusion-repr/src/plan_nodes/projection.rs
```diff
@@ -26,3 +26,57 @@ define_plan_node!(
         { 1, exprs: ExprList }
     ]
 );
+
+pub struct ProjectionMapping {
+    forward: Vec<usize>,
+    _backward: Vec<Option<usize>>,
+}
+
+impl ProjectionMapping {
```
Sweetsuro, line 47: Maybe add a comment here specifying how projection mapping works/its intention?

optd-datafusion-repr/src/testing.rs
```diff
@@ -0,0 +1,27 @@
...
+use self::tpch_catalog::TpchCatalog;
+
+/// Create a "dummy" optimizer preloaded with the TPC-H catalog for testing
+/// Note: Only provides the schema property currently
+pub fn new_test_optimizer(
+    rule: Arc<dyn Rule<OptRelNodeTyp, HeuristicsOptimizer<OptRelNodeTyp>>>,
+) -> HeuristicsOptimizer<OptRelNodeTyp> {
+    let dummy_catalog = Arc::new(TpchCatalog);
+
+    HeuristicsOptimizer::new_with_rules(
```
Sweetsuro, line 22: Switch to cascades optimizer?

optd-datafusion-repr/src/testing/dummy_cost.rs
```diff
@@ -0,0 +1,37 @@
+use crate::plan_nodes::OptRelNodeTyp;
```
Sweetsuro, line 1: Is optd still able to select a winner if all costs are the same?

optd-datafusion-repr/src/rules/filter_pushdown.rs
```diff
@@ -0,0 +1,804 @@
...
+
+#[derive(Debug, Clone, Copy)]
+enum JoinCondDependency {
+    Left,
+    Right,
+    Both,
+    None,
+}
+
+fn determine_join_cond_dep(
```
Sweetsuro, line 52: Add a comment for what this is used for maybe?

optd-datafusion-repr/src/rules/filter_pushdown.rs
```diff
@@ -0,0 +1,804 @@
...
+define_rule!(
+    FilterProjectTransposeRule,
+    apply_filter_project_transpose,
+    (Filter, (Projection, child, [exprs]), [cond])
+);
+
+/// Datafusion only pushes filter past project when the project does not contain
+/// volatile (i.e. non-deterministic) expressions that are present in the filter
+/// Calcite only checks if the projection contains a windowing calculation
+/// We check neither of those things and do it always (which may be wrong)
```
Sweetsuro, line 134: Add this to some kind of issue list? Along with the missing pull up rules

optd-datafusion-repr/src/rules/filter_pushdown.rs
```diff
@@ -0,0 +1,804 @@
...
+        child_b,
+        join_cond,
+        cond,
+    )
+}
+
+/// Cases:
+/// - Push down to the left child (only involves keys from the left child)
+/// - Push down to the right child (only involves keys from the right child)
+/// - Push into the join condition (involves keys from both children)
```
Sweetsuro, line 234: For something like 
filter cond = a.0=1 OR b.1=1 
or filter cond = (a.0=1 AND a.1) OR b.1=1

when there are no ANDs across tables, logically I think you'd still want to push the separated filters pass the join cond

optd-datafusion-repr/src/rules/filter_pushdown.rs
```diff
@@ -0,0 +1,804 @@
...
+            LogOpExpr::new(LogOpType::And, ExprList::new(push_conds)).into_expr(),
+        )
+        .into_plan_node()
+    } else {
+        child
+    };
+
+    let new_agg = LogicalAgg::new(new_child, exprs, groups);
+
+    let new_filter = LogicalFilter::new(
```
Sweetsuro: Extra filter node on the outside of agg here. Also can flatten the LogOpExpr

optd-datafusion-repr/src/rules/filter_pushdown.rs
```diff
@@ -0,0 +1,804 @@
...
+        } else {
+            keep_conds.push(expr);
+        }
+    };
+    categorize_conds(categorization_fn, Expr::from_rel_node(cond.into()).unwrap());
+
+    let new_child = if !push_conds.is_empty() {
+        LogicalFilter::new(
+            child,
+            LogOpExpr::new(LogOpType::And, ExprList::new(push_conds)).into_expr(),
```
Sweetsuro: flatten the LogOpExpr

Sweetsuro (changes requested):
Left a couple comments that should probably be addressed

optd-datafusion-repr/src/lib.rs
jurplel: Moved the rules. We can add them to some list of potential future rules

optd-datafusion-repr/src/plan_nodes/projection.rs
jurplel: There was already an example, but i fleshed it out a bit more. hopefully better.

optd-datafusion-repr/src/rules/filter_pushdown.rs
jurplel: This is going to have to come with support for data about the volatility of an expression, which i don't think is on the road map right now. Where should I note this down?

optd-datafusion-repr/src/testing/dummy_cost.rs
jurplel: You're right, and this is blocking switching to cascades I think. It is unable to find a winner when I switch to cascades.

Making a dummy cost model with actual numbers is kinda non-trivial i think

optd-datafusion-repr/src/rules/filter_pushdown.rs
```diff
@@ -0,0 +1,813 @@
...
+    Left,
+    Right,
+    Both,
+    None,
+}
+
+/// Given a list of expressions (presumably a flattened tree), determine
+/// if the expression is dependent on the left child, the right child, both,
+/// or neither, by analyzing which columnrefs are used in the expressions.
+fn determine_join_cond_dep(
```
yliang412, line 52: potentially useful for other rules, we can make it pub later.

optd-datafusion-repr/src/rules/filter_pushdown.rs
```diff
@@ -0,0 +1,813 @@
...
+    }
+    match (left_col, right_col) {
+        (true, true) => JoinCondDependency::Both,
+        (true, false) => JoinCondDependency::Left,
+        (false, true) => JoinCondDependency::Right,
+        (false, false) => JoinCondDependency::None,
+    }
+}
+
+/// Do not call directly
```
yliang412: Could define the helper nested in the actual function

optd-datafusion-repr/src/rules/filter_pushdown.rs
```diff
@@ -0,0 +1,813 @@
...
+}
+
+define_rule!(
+    FilterAggTransposeRule,
+    apply_filter_agg_transpose,
+    (Filter, (Agg, child, [exprs], [groups]), [cond])
+);
+
+/// Filter is commutable past aggregations when the filter condition only
+/// involves the group by columns.
```
yliang412: What are we doing if the filter condition involves the agg columns?

yliang412: nvm, you only push the filters that can be pushed, the other ones are kept. Could you maybe document this? I feel although everything is called `FilterXXXTransposeRule`, what they actually do are kinda different.

yliang412 (approved):
Overall LGTM. Feel free to merge after addressing the comments (or do it in a separate PR)

optd-datafusion-repr/src/testing.rs
jurplel: #152

optd-datafusion-repr/src/rules/filter_pushdown.rs
jurplel: noted in #153

jurplel:
@Sweetsuro all comments finally addressed—approve and i will merge it!

Sweetsuro (approved):
LGTM! Maybe update the PR description before merging

PR #141, opened 2024-03-27
feat: display logical plan after heuristic

yliang412:
It might be helpful to see what the plan looks like after the heuristics phase. It might help us to identify things that are not handled in the system and help us make design decisions forward on implementing rules.


jurplel (approved):
lgtm!

PR #142, opened 2024-03-29
fix: agg now giving row cnt of 1

patrickhyw:
**Summary**: see title

**Demo**:
We now get perfect q-error on queries 14 and 17:
Screenshot 2024-03-29 at 18 51 14


skyzh:
But that could also be a group aggregation?

patrickhyw:
Yup, good catch, that's why I closed the PR.

Issue #127, opened 2024-03-22
Tracking: parity with Postgres for TPC-H cardinality estimations


patrickhyw:
### Notes
* Sometimes Postgres does really bad (even worse than our magic numbers!). However, the goal right now is simply to match Postgres, not to match the truecard. When I say **fix**, I mean **match Postgres**.
  * This is because we know exactly what we need to do to match Postgres but we don't know what we need to do to match the truecard.
* Experiments ran with scale factor 1.0, seed 15721
* If no other PR is mentioned, then the query was run based on #132 

### Queries
- [ ] Q1
  - Not running. See #68 
- [x] Q2 (already matching in #132)
- [x] Q3: truecard=10, pgcard=10, dfcard=10
  - Fixed by #138 
- [ ] Q4
  - Not running. See #68 
- [x] Q5: truecard=5, pgcard=25, dfcard=25
  - Fixed by #144 
- [x] Q6: truecard=1, pgcard=1, dfcard=1
  - #143 revealed the problem here
  - Fixed by #144 
- [ ] Q7: truecard=4, pgcard=6119, dfcard=125000
  - **Fixing join predicates** and **fixing multi-dim group by** would definitely help with this, but it's not clear whether it would completely fix it.
  - #145 changed dfcard from 1 to 125000
- [ ] Q8: truecard=2, pgcard=2406, dfcard=200
  - **Fixing single-dim group by** and **pulling expressions up to the group by** should fix this. Postgres identifies that the group by is done on `EXTRACT(year FROM orders.o_orderdate)` and it simply uses the N-Distinct of orders.o_orderdate as the cardinality of the query.
- [ ] Q9: truecard=175, pgcard=60150, dfcard=5000
  - **Fixing single-dim group by** and **pulling expressions up to the group by** should fix this. When you get rid of `p_name like '%forest'` and just use `o_orderdate as o_year`, you get exactly 60150 rows.
  - #145 changed dfcard from 25 to 5000
- [x] Q10: truecard=20, pgcard=20, dfcard=20
    - Fixed by #138 and #143
- [ ] Q11: truecard=869, pgcard=10667, dfcard=67936
  - #145 changed dfcard from 1 to 67936
  - I'm not sure how Postgres gets to 10667.
- [x] Q12: truecard=2, pgcard=7, dfcard=7
  - Fixed by #143 and #144 
- [x] Q13: truecard=42, pgcard=200, dfcard=200
  - Fixed by #145 
[...]

PR #143, opened 2024-03-30
feat: using proper magic numbers in various edge cases

patrickhyw:
**Summary**: Using magic numbers from Postgres in various selectivity edge cases.

**Demo**:

Different (unfortunately worse) q-error on TPC-H SF1. See #127 for per-query details on how this PR affects q-error.

Screenshot 2024-03-30 at 11 27 24

**Details**:
* Fixed the cardinality of Q10!
* `INVALID_SEL` is **no longer used** at all during cardtest. It is still used during plannertest as some plannertests use the optd optimizer instead of the datafusion logical optimizer. This can be checked by replacing all instances of `INVALID_SEL` with a `panic!()` and seeing that cardtest still runs.
* Using magic number from Postgres for `LIKE`.
* Using magic number from Postgres for equality with various complex expressions.
* Using magic number from Postgres for range comparison with various complex expressions.
* Replaced `INVALID_SEL` with `panic!()` and `unreachable!()` statements in places where it makes sense.


Gun9niR (approved):
LGTM

PR #144, opened 2024-03-30
feat: add cost estimation for agg

Gun9niR:
Compute cost for aggregation. The cardinality is computed as the product of n-distinct of all the group-by columns. If there's no group by column, the output cardinality is just 1.

This should fix the cardinality parity between postgres for Q14 and Q17. It also leads to a better join order in Q11, since aggregation is the child of a join.

## Misc

- Add planner test for Q11.
- Fixes Q14 and Q17.
- Next step is to support n-distinct for string.
- We may change to multi-dimension n-distinct after it's integrated.

### DIFF
--- a/optd-datafusion-repr/src/cost/base_cost.rs
+++ b/optd-datafusion-repr/src/cost/base_cost.rs
@@ -324,7 +324,7 @@ const DEFAULT_INEQ_SEL: f64 = 0.3333333333333333;
 // Default selectivity estimate for pattern-match operators such as LIKE
 const DEFAULT_MATCH_SEL: f64 = 0.005;
 // Default n-distinct estimate for derived columns or columns lacking statistics
-const DEFAULT_N_DISTINCT: u64 = 1;
+const DEFAULT_N_DISTINCT: u64 = 200;
 
 const INVALID_SEL: f64 = 0.01;
 
--- a/optd-datafusion-repr/src/cost/base_cost.rs
+++ b/optd-datafusion-repr/src/cost/base_cost.rs
@@ -323,6 +323,8 @@ const DEFAULT_EQ_SEL: f64 = 0.005;
 const DEFAULT_INEQ_SEL: f64 = 0.3333333333333333;
 // Default selectivity estimate for pattern-match operators such as LIKE
 const DEFAULT_MATCH_SEL: f64 = 0.005;
+// Default selectivity if we have no information
+const DEFAULT_UNK_SEL: f64 = 0.005;
 // Default n-distinct estimate for derived columns or columns lacking statistics
 const DEFAULT_N_DISTINCT: u64 = 200;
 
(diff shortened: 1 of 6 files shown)


optd-datafusion-repr/src/cost/base_cost.rs
```diff
@@ -323,6 +323,8 @@ const DEFAULT_EQ_SEL: f64 = 0.005;
 const DEFAULT_INEQ_SEL: f64 = 0.3333333333333333;
 // Default selectivity estimate for pattern-match operators such as LIKE
 const DEFAULT_MATCH_SEL: f64 = 0.005;
+// Default n-distinct estimate for derived columns or columns lacking statistics
+const DEFAULT_N_DISTINCT: u64 = 1;
```
patrickhyw: Postgres's default is 200, let's change it to that.
Screenshot 2024-03-30 at 18 51 02

optd-datafusion-repr/src/cost/base_cost.rs
```diff
@@ -544,6 +551,65 @@ impl<M: MostCommonValues, D: Distribution> OptCostModel<M, D> {
...
+                                ColumnRef::Derived => DEFAULT_N_DISTINCT,
+                                _ => panic!("GROUP BY base table column ref must either be derived or base table"),
+                            })
+                            .product::<u64>() as f64
+                    }
+                } else {
+                    panic!("compute_cost() should not be called if optimizer is None")
+                }
+            }
+            None => panic!("compute_cost() should not be called if context is None"),
```
patrickhyw: I think we should do `(row_cnt * DEFAULT_UNK_SEL).max(1.0)` instead of `panic!()` if the context/optimizer doesn't exist. It doesn't seem like a case that warrants panicking. I removed all panics in my join selectivity branch.

optd-datafusion-repr/src/properties/schema.rs
```diff
@@ -96,6 +96,19 @@ impl PropertyBuilder<OptRelNodeTyp> for SchemaPropertyBuilder {
                     children.len()
                 ],
             },
+            OptRelNodeTyp::Agg => {
+                let mut schema = children[1].clone();
+                let schema2 = children[2].clone();
```
patrickhyw: Maybe make variables for `children[1]` and `children[2]` with names for what they are (group-by and aggregations) so that the code is more readable?

optd-datafusion-repr/src/properties/schema.rs
```diff
@@ -96,6 +96,19 @@ impl PropertyBuilder<OptRelNodeTyp> for SchemaPropertyBuilder {
...
+                schema.fields.extend(schema2.fields);
+                schema
+            }
+            OptRelNodeTyp::Func(FuncType::Agg(_)) => Schema {
+                fields: vec![Field {
+                    name: DEFAULT_NAME.to_string(),
+                    typ: ConstantType::Any,
+                    nullable: true,
```
patrickhyw: Per our discussion, add a comment above the `Field` struct and make this `Field::placeholder()`

patrickhyw (approved):
LGTM other than small changes

patrickhyw:
Oh one more thing, could you update the q-errors in #127 after each PR? This way, we always know which PR caused which changes to the q-errors? I noticed that you marked Q14 as fixed but didn't update `dfcard`

PR #147, opened 2024-04-02
Fix: fix for heuristic rule wrapper

AveryQi115:
The new expr returned by heuristic rules are not in the original group, which means it never got an OptimizeExpressionTask with exploring as false (OptimizeExpressionTask with exploring=False only be called in OptimizeGroup), it should evoke an OptimizeExpressionTask with exploring=false for itself to apply all the transform rules and implementation rules for itself.

Besides, the pr adds checks for whether the original expr equals to the output expr for heuristic rule. If that's the case, it should prompt an error as this breaks the heuristic rule's definition. (Silently accepting it will mark the input expr being as a dead end and there's no more new exprs to replace it)

### DIFF
--- a/optd-core/src/cascades/memo.rs
+++ b/optd-core/src/cascades/memo.rs
@@ -285,13 +285,19 @@ impl<T: RelNodeTyp> Memo<T> {
             };
 
             // if the new expr already in the memo table, merge the group and remove old expr
-            if let Some(&expr_id) = self.expr_node_to_expr_id.get(&memo_node) {
-                let group_id = self.get_group_id_of_expr_id(expr_id);
+            if let Some(&new_expr_id) = self.expr_node_to_expr_id.get(&memo_node) {
+                if new_expr_id == expr_id {
+                    // This is not acceptable, as it means the expr returned by a heuristic rule is exactly
+                    // the same as the original expr, which should not happen
+                    // TODO: we can silently ignore this case without marking the original one as a deadend
+                    // But the rule creators should follow the definition of the heuristic rule
+                    // and return an empty vec if their rule does not do the real transformation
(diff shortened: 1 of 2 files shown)


Sweetsuro (approved):
LGTM - thanks for helping me find the bug!

skyzh:
did we consider merge two groups of the same ID in cascades? maybe we also need to fix that logic.

AveryQi115:
> did we consider merge two groups of the same ID in cascades? maybe we also need to fix that logic.

I've thought about this case. But I don't think we need for this one. As the input two groups for merge group are not newly generated by apply_rule task. They exist before the apply_rule task. So there must be optimizeGroup task for them already in the task list. And the exprs in the two groups then have been evoked OptimizeExpression task, even they are moved from one group to another group?

AveryQi115:
> > did we consider merge two groups of the same ID in cascades? maybe we also need to fix that logic.
> 
> I've thought about this case. But I don't think we need for this one. As the input two groups for merge group are not newly generated by apply_rule task. They exist before the apply_rule task. So there must be optimizeGroup task for them already in the task list. And the exprs in the two groups then have been evoked OptimizeExpression task, even they are moved from one group to another group?

But that is based on the assumption that the rules returning a group cannot create new groups in rules logic, they can only return groups that already existed(for example, a rule is extracting filter node and its child and only returns its child node and eliminate the filter)

AveryQi115:
Okay, we probably need that fix for merge group too then. The optimizeGroup task is only evoked for the root relnode, and other optimzieGroup task are called from optimizeInput(when it wants to optimize its children group). If a child group after some rule application is merged to a top level group which is already called OptimizeGroup task, the optimize group task needs to be optimized again for its newly added expressions.

PR #154, opened 2024-04-11
feat: implement filter selectivity for `InListExpr`

Gun9niR:
Only support the form of `colA = (1, 2, 3)`, where `colA` is a base table column and `(1, 2, 3)` consists of all constants. The selectivity is computed by **adding the selectivity of individual elements together**.


patrickhyw (approved):
LGTM

PR #156, opened 2024-04-12
refactor: split base_cost.rs into multiple files

Gun9niR:
Move cost model code to separate files.

- Statistics generation -> `src/cost/base_cost/stats.rs`
- Filter selectivity -> `src/cost/base_cost/filter.rs`
- Join selectivity -> `src/cost/base_cost/join.rs`
- Limit cardinality -> `src/cost/base_cost/limit.rs`
- Agg cardinality -> `src/cost/base_cost/agg.rs`


optd-datafusion-repr/src/cost/stats.rs
```diff
@@ -1,7 +1,42 @@
-use optd_core::rel_node::Value;
```
patrickhyw, line 1: Should this be in `base_cost/`?

Gun9niR: Fixed

patrickhyw (approved):
LGTM

PR #159, opened 2024-04-14
feat: can load files > 1GB into postgres

patrickhyw:
**Summary**: Can now load files \> 1GB in size by chunking files into 2MB chunks for `COPY FROM STDIN`.

**Demo**:
JOB's `cast_info.csv` is 1.3GB and has 36244344 rows (image 1). All 36244344 are loaded successfully (image 2).
Screenshot 2024-04-14 at 18 06 22
Screenshot 2024-04-14 at 18 06 36

**Details**:
* Postgres does not allow statements that exceed 1GB in size. Since we previously sent the entire file in a single statement, this caused Postgres to cancel the transaction.
* This is fixed by chunking files into 2MB chunks and doing a separate `COPY FROM STDIN` statement for each.
* We also solved another issue in the meantime: we can now load files that would not fit in memory.

### DIFF
--- a/optd-perftest/src/main.rs
+++ b/optd-perftest/src/main.rs
@@ -1,9 +1,9 @@
 use clap::{Parser, Subcommand, ValueEnum};
+use optd_perftest::benchmark::Benchmark;
 use optd_perftest::cardtest;
 use optd_perftest::job::JobConfig;
 use optd_perftest::shell;
 use optd_perftest::tpch::{TpchConfig, TPCH_KIT_POSTGRES};
-use optd_perftest::benchmark::Benchmark;
 use prettytable::{format, Table};
 use std::fs;
 use std::path::Path;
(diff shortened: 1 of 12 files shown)


optd-perftest/src/main.rs
```diff
@@ -57,11 +71,112 @@ enum Commands {
...
+        BenchmarkName::Tpch => Benchmark::Tpch(TpchConfig {
+            dbms: String::from(TPCH_KIT_POSTGRES),
+            scale_factor,
+            seed,
+            query_ids: query_ids.clone(),
+        }),
+        BenchmarkName::Job => Benchmark::Job(JobConfig {
+            query_ids: query_ids.clone(),
+        }),
+    };
```
Gun9niR, line 102: It's probably better to build the `benchmark` parameter outside of this function.

Gun9niR (approved):
Rest LGTM

optd-perftest/src/main.rs
patrickhyw: I think of `cardtest()` as containing all the functionality of handling cardinality testing, which includes processing the input from raw command line arguments, which is why I build it inside the function.

patrickhyw: I want to keep code out of `main()` because it would get too cluttered as we added more types of tests.

PR #160, opened 2024-04-15
feat: multi column stats

AlSchlo:
This PR adds support for multi column statistics & minor bug-fixes.

### DIFF
--- a/optd-datafusion-repr/src/cost/base_cost/stats.rs
+++ b/optd-datafusion-repr/src/cost/base_cost/stats.rs
@@ -134,13 +134,20 @@ impl<M: MostCommonValues, D: Distribution> ColumnCombValueStats<M, D> {
 
 #[serde_with::serde_as]
 #[derive(Serialize, Deserialize, Debug)]
-pub struct TableStats<M: MostCommonValues + Serialize + DeserializeOwned, D: Distribution + Serialize + DeserializeOwned> {
+pub struct TableStats<
+    M: MostCommonValues + Serialize + DeserializeOwned,
+    D: Distribution + Serialize + DeserializeOwned,
+> {
     pub row_cnt: usize,
     #[serde_as(as = "HashMap<serde_with::json::JsonString, _>")]
     pub column_comb_stats: HashMap<ColumnsIdx, ColumnCombValueStats<M, D>>,
 }
 
-impl<M: MostCommonValues + Serialize + DeserializeOwned, D: Distribution + Serialize + DeserializeOwned> TableStats<M, D> {
+impl<
+        M: MostCommonValues + Serialize + DeserializeOwned,
+        D: Distribution + Serialize + DeserializeOwned,
+    > TableStats<M, D>
+{
     pub fn new(
         row_cnt: usize,
         column_comb_stats: HashMap<ColumnsIdx, ColumnCombValueStats<M, D>>,
--- a/optd-datafusion-repr/src/cost/base_cost/stats.rs
+++ b/optd-datafusion-repr/src/cost/base_cost/stats.rs
@@ -1,6 +1,6 @@
 use std::{collections::HashMap, sync::Arc};
 
-use arrow_schema::{ArrowError, DataType};
+use arrow_schema::{ArrowError, DataType, SchemaRef};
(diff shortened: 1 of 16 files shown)


optd-datafusion-repr/src/cost/base_cost/stats.rs
```diff
@@ -249,6 +353,105 @@ impl PerTableStats<Counter<Value>, TDigest> {
...
+    }
+
+    fn get_typed_rows(
+        cols: Vec<Arc<dyn Array>>,
+        col_types: &[DataType],
+    ) -> Vec<Vec<Option<Value>>> {
+        let mut rows = Vec::<Vec<Option<Value>>>::new();
+        for (col, col_type) in cols.iter().zip(col_types.iter()) {
+            for (idx, val) in Self::to_typed_column(col, col_type).iter().enumerate() {
+                rows[idx].push(val.clone());
```
Gun9niR: This will cause out-of-bound access.

optd-datafusion-repr/src/cost/base_cost/stats.rs
```diff
@@ -249,6 +353,105 @@ impl PerTableStats<Counter<Value>, TDigest> {
...
+        col_types: &[DataType],
+        mg: &mut MisraGries<Vec<Option<Value>>>,
+        hll: &mut HyperLogLog,
+        null_count: &mut i32,
+    ) {
+        let rows = Self::get_typed_rows(cols, col_types);
+
+        let (fully_null_rows, non_fully_null_rows): (Vec<_>, Vec<_>) = rows
+            .into_iter()
+            .partition(|row| row.iter().all(|v| v.is_none()));
```
Gun9niR: This operation might be very costly, as all the rows are copied once more.

optd-datafusion-repr/src/cost/base_cost/stats.rs
```diff
@@ -230,9 +249,94 @@ impl PerTableStats<Counter<Value>, TDigest> {
...
+        cols
+    }
+
+    pub fn extend_with_multi_column_stats<
+        I: IntoIterator<Item = Result<RecordBatch, ArrowError>>,
+    >(
+        &mut self,
+        batch_iter_builder: impl Fn() -> anyhow::Result<RecordBatchIterator<I>>,
+        combinations: Vec<Vec<usize>>,
+    ) -> anyhow::Result<()> {
```
Gun9niR: How do you plan to use this function? Looks like each invocation is two more scans.

optd-datafusion-repr/src/cost/base_cost/stats.rs
```diff
@@ -103,7 +105,7 @@ pub struct PerColumnStats<M: MostCommonValues, D: Distribution> {
 
     // distribution _does not_ include the values in mcvs
     // distribution _does not_ include nulls
-    pub distr: D,
+    pub distr: Option<D>,
```
Gun9niR: Should we separate `PerColumnStats` and `CombColumnStats`? Just in case they diverge further.

AlSchlo: yeah, probably. also has a better naming

optd-datafusion-repr/src/cost/base_cost/stats.rs
AlSchlo: each invocation is len(combinations) x 2 scans, right now, for simplicity
it can be made x2 scans only

actually, i will fully refactor all this before merging anything

optd-datafusion-repr/src/cost/base_cost/stats.rs
AlSchlo: right... well the previous operation had a "flatten" and a "collect". that also copied stuff no?

AlSchlo: i coudn't find a functional way to count while filtering

optd-datafusion-repr/src/cost/base_cost/stats.rs
AlSchlo: true, i'll add some tests

PR #161, opened 2024-04-15
feat: JOB cardinality testing

patrickhyw:
**Summary**: Basic JOB cardinality testing done with Postgres cost estimation, DataFusion stats building, and optd cost estimation for queries 1a and 1b.

**Demo**:
Working for queries 1a and 1b.
Screenshot 2024-04-15 at 12 07 43

**Details**:
* HLL and MCVs currently cause issues that will be fixed in future PRs, so I create stats for JOB right now with a special `from_record_batches_job()`.
* I only did queries 1a and 1b because figuring out which queries work is a separate problem (and should wait until we have the stats completed).


Gun9niR (approved):
LGTM

PR #162, opened 2024-04-17
feat: projection merge and projection filter transpose rule

Sweetsuro:
This PR implements a part of the projection transpose series of rules. It also includes a fair amount of refactoring.

### Projection Merge Rule
- This rule matches on two projection nodes and combines the two nodes into one.
- It is added to the heuristic optimizer pass before the cascades optimizer. In the future, it should also be added to a pass after the cascades optimizer.

### Projection Filter Transpose Rule
- This rule matches pushes a projection node passed a filter node. If the filter node contains columns that are not in this projection node, the top most projection node is also kept.
- It is added as a cascades rule.

### Refactoring
Relevant functions for projection transpose rules can be found in `project_transpose_common.rs`.  Rules are implemented in separate files as a part of the `projection_transpose` module rather than in all one file. Similarly, `FilterProjectTransposeRule` and `ProjectionPullUpJoin` were moved into this module.

### Testing
Unit tests using the dummy heuristic optimizer were implemented.


skyzh (approved):
Generally LGTM, good work :)

jurplel (approved):
LGTM, nice work on testing. Sorry for the delay.

PR #164, opened 2024-04-19
fix: reference the correct column stats for nlj and hash join

Gun9niR:
The indices of `ColumnRefExpr`s in hash join columns are with respect to left or right tables. For example, 

image

The plan in this figure means join on 0th column of the left table and 3rd column of the right table.

But for NLJ, the column indices are with respect to the join output. For instance, if the condition is `#0 = #5` and the left table has 3 columns, the condition actually means join on 0th column of the left table and 3rd column of the right.

### DIFF
--- a/optd-datafusion-repr/src/cost/base_cost/join.rs
+++ b/optd-datafusion-repr/src/cost/base_cost/join.rs
@@ -102,6 +102,7 @@ impl<
     }
 
     /// A wrapper to convert the join keys to the format expected by get_join_selectivity_core()
+    #[allow(clippy::too_many_arguments)]
     fn get_join_selectivity_from_keys(
         &self,
         join_typ: JoinType,
@@ -139,6 +140,7 @@ impl<
     }
 
     /// The core logic of join selectivity which assumes we've already separated the expression into the on conditions and the filters
+    #[allow(clippy::too_many_arguments)]
     fn get_join_selectivity_core(
         &self,
         join_typ: JoinType,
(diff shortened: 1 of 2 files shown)


optd-datafusion-repr/src/cost/base_cost/join.rs
```diff
@@ -86,6 +89,7 @@ impl<
                 &column_refs,
                 row_cnt_1,
                 row_cnt_2,
+                left_col_cnt,
```
patrickhyw, line 92: Why is this called `left_col_cnt` here but `right_col_ref_offset` in the function signature?

optd-datafusion-repr/src/cost/base_cost/join.rs
```diff
@@ -106,6 +111,7 @@ impl<
         column_refs: &GroupColumnRefs,
         left_row_cnt: f64,
         right_row_cnt: f64,
+        right_col_ref_offset: usize,
```
patrickhyw: What does this do? Why do we only need `right_col_ref_offset` and not `left_col_ref_offset`? Could you add comments explaining these?

patrickhyw (changes requested):

optd-datafusion-repr/src/cost/base_cost/join.rs
Gun9niR: https://github.com/cmu-db/optd/blob/4e3ede01f4666543d86fbb1befc3129248abe545/optd-datafusion-repr/src/cost/base_cost/join.rs#L143-L148

optd-datafusion-repr/src/cost/base_cost/join.rs
Gun9niR: Because for hash join, the indices of `ColumnRef`s on the right side should be offset by the number of columns in the left table to get the correct `GroupColumnRefs`, whereas for NLJ the indices can be used as is.

patrickhyw (approved):
LGTM

PR #165, opened 2024-04-20
feat: refine like selectivity computation

Gun9niR:
See `get_like_selectivity` for details. This fix significantly narrows the selectivity gap of `LIKE` between optd and Postgres on JOB.

https://github.com/cmu-db/optd/blob/4fae01af3cbc723be5b6a85e1fae8bc31c1fa297/optd-datafusion-repr/src/cost/base_cost/filter.rs#L347-L357

### DIFF
--- a/optd-datafusion-repr/src/cost/base_cost/filter.rs
+++ b/optd-datafusion-repr/src/cost/base_cost/filter.rs
@@ -1,6 +1,5 @@
 use std::ops::Bound;
 
-use datafusion::arrow::{array::StringArray, compute::like};
 use optd_core::{
     cascades::{CascadesOptimizer, RelNodeContext},
     cost::Cost,
@@ -23,6 +22,8 @@ use crate::{
 use super::{
     stats::ColumnCombValue, OptCostModel, DEFAULT_EQ_SEL, DEFAULT_INEQ_SEL, DEFAULT_UNK_SEL,
 };
+
+mod like;
 
 impl<
         M: MostCommonValues + Serialize + DeserializeOwned,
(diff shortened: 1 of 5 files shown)


optd-datafusion-repr/src/cost/base_cost/filter.rs
```diff
@@ -340,6 +344,87 @@ impl<
...
+    /// population and then add up data for any MCV values. If the histogram is large enough, it just uses the number
+    /// of matches in the histogram, otherwise it estimates the fixed prefix and remainder of pattern separately and
+    /// combine them.
+    ///
+    /// Our approach is simpler and less selective. Firstly, we don't use histogram. The selectivity is composed of MCV
+    /// frequency and non-MCV selectivity. MCV frequency is computed by adding up frequencies of MCVs that match the
+    /// pattern. Non-MCV  selectivity is computed in the same way that Postgres computes selectivity for the wildcard
+    /// part of the pattern.
+    fn get_like_selectivity(&self, like_expr: &LikeExpr, column_refs: &GroupColumnRefs) -> f64 {
+        const FULL_WILDCARD_SEL: f64 = 5.0;
```
patrickhyw: How could selectivity be > 1.0?

optd-datafusion-repr/src/cost/base_cost/filter.rs
```diff
@@ -340,6 +344,87 @@ impl<
...
+                )
+            } else {
+                (0.0, 0.0)
+            };
+
+            // Postgres clamps the result after histogram and before MCV. See Postgres `patternsel_common`.
+            let result = (non_mcv_sel + mcv_freq * (1.0 - null_frac)).clamp(0.0001, 0.9999);
+
+            if like_expr.negated() {
+                1.0 - result - null_frac
```
patrickhyw: Nice catch with the null frac

optd-datafusion-repr/src/cost/base_cost/filter.rs
```diff
@@ -340,6 +344,87 @@ impl<
...
+    /// The logic is somewhat similar to Postgres but different. Postgres first estimates the histogram part of the
+    /// population and then add up data for any MCV values. If the histogram is large enough, it just uses the number
+    /// of matches in the histogram, otherwise it estimates the fixed prefix and remainder of pattern separately and
+    /// combine them.
+    ///
+    /// Our approach is simpler and less selective. Firstly, we don't use histogram. The selectivity is composed of MCV
+    /// frequency and non-MCV selectivity. MCV frequency is computed by adding up frequencies of MCVs that match the
+    /// pattern. Non-MCV  selectivity is computed in the same way that Postgres computes selectivity for the wildcard
+    /// part of the pattern.
+    fn get_like_selectivity(&self, like_expr: &LikeExpr, column_refs: &GroupColumnRefs) -> f64 {
```
patrickhyw: Could you add some unit tests testing MCV, non-MCV, and nulls?

patrickhyw (changes requested):

optd-datafusion-repr/src/cost/base_cost/filter.rs
Gun9niR: Actually in practice it makes no difference since arrow treats NULL strings as empty instead of NULL.

optd-datafusion-repr/src/cost/base_cost/filter.rs
Gun9niR: The idea is that WILDCARD makes the pattern "less selective". This constant is used when estimating the pattern's selectivity char-by-char. Plus we clamp the final result to [0.0001, 0.9999] same as Postgres.

optd-datafusion-repr/src/cost/base_cost/filter.rs
Gun9niR: Added

optd-datafusion-repr/src/cost/base_cost/filter.rs
patrickhyw: Sounds good. Could you add that as a comment then so people don't think it's a mistake?

patrickhyw (approved):
LGTM other than my one comment

PR #166, opened 2024-04-21
fix: found which queries work for JOB with automated script

patrickhyw:
**Summary**: figured out the definitive list of 98 / 113 queries which are currently working in optd.

**Details**:
* "Work" is with DataFusion's logical optimizer, not optd's.
* Also wrote an automated script, `which_queries_work.sh`, which can automatically update this list as we get more queries to work.
* `which_queries_work.sh` also works with TPC-H as well


Gun9niR (approved):

AlSchlo (approved):
cool bashing skilzs

PR #167, opened 2024-04-24
optimize: speed up stat gen by factor x15

AlSchlo:
Made the stat generation faster using Rayon's thread pool.

Various improvements, such as;
- slightly less copying
- optimized MG
- process in parallel using Rayon, taking advantage of fold/reduce
- moved to Parquet (WIP to convert into Parquet automatically)
- read Parquet in parallel (one reader for each row-group), this granularity is sufficient for big enough datasets

This takes 30s for JOB 1D stats on my computer, vs 7:30min before.
Postgres takes 1:30min loading, and 22s for the stat gen. So we beat it, depending on how we view it.

On "real" datacenter hardware (i.e. 512 cores), we would **crush** it, we'll test that soon.

Finally coming together :-)

### DIFF
--- a/optd-datafusion-repr/src/cost/base_cost/stats.rs
+++ b/optd-datafusion-repr/src/cost/base_cost/stats.rs
@@ -1,13 +1,12 @@
-use std::{
-    collections::HashMap,
-    sync::{mpsc::Receiver, Arc},
-    thread::JoinHandle,
-};
+use std::{collections::HashMap, sync::Arc};
 
-use arrow_schema::{ArrowError, DataType, Schema, SchemaRef};
-use datafusion::arrow::array::{
-    Array, BooleanArray, Date32Array, Float32Array, Int16Array, Int32Array, Int8Array, RecordBatch,
-    StringArray, UInt16Array, UInt32Array, UInt8Array,
+use arrow_schema::{DataType, Schema, SchemaRef};
+use datafusion::{
+    arrow::array::{
+        Array, BooleanArray, Date32Array, Float32Array, Float64Array, Int16Array, Int32Array,
+        Int8Array, RecordBatch, StringArray, UInt16Array, UInt32Array, UInt8Array,
+    },
+    parquet::arrow::arrow_reader::ParquetRecordBatchReader,
 };
 use itertools::Itertools;
 use optd_core::rel_node::{SerializableOrderedF64, Value};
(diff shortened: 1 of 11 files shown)


optd-gungnir/src/stats/misragries.rs
```diff
@@ -5,11 +5,11 @@
...
 //! https://people.csail.mit.edu/rrw/6.045-2017/encalgs-mg.pdf
 
-use std::{cmp::min, collections::HashMap, hash::Hash};
-
+use hashbrown::HashMap;
 use itertools::Itertools;
+use std::{cmp::min, hash::Hash};
 
-pub const DEFAULT_K_TO_TRACK: u16 = 1000;
+pub const DEFAULT_K_TO_TRACK: u16 = 200;
```
patrickhyw, line 12: This may reduce our accuracy. Do you know what effect this has on q-error?

optd-datafusion-repr/src/cost/base_cost/stats.rs
```diff
@@ -339,50 +394,95 @@ impl TableStats<Counter<ColumnCombValue>, TDigest<Value>> {
...
-                StatType::Full => Some(TDigest::new(tdigest::DEFAULT_COMPRESSION)),
-                StatType::Partial => None,
-            })
-            .collect_vec();
-        let mut cnts = mgs
-            .iter()
-            .map(|mg| {
-                let mfk = mg.most_frequent_keys().into_iter().cloned().collect_vec();
-                Counter::new(&mfk)
+        // TODO(Alexis): This materialization is OK as JOB only takes 1GB, but should be made in parallel...
```
patrickhyw: I was thinking of doing a TPC-H SF10 test. Would this code still work at SF10 (10GB)?

optd-datafusion-repr/src/cost/base_cost/stats.rs
```diff
@@ -339,50 +394,95 @@ impl TableStats<Counter<ColumnCombValue>, TDigest<Value>> {
...
+                let mut local_stats = local_stats?;
+
+                match batch {
+                    Ok(batch) => {
+                        let (hlls, mgs, null_cnts) = &mut local_stats;
+                        let comb = Self::get_column_combs(batch, &comb_stat_types);
+                        Self::generate_partial_stats(&comb, mgs, hlls, null_cnts);
+                        Ok(local_stats)
+                    }
+                    Err(_) => todo!(), // TODO(Alexis): Could not satisfy the type checker otherwise, but never happens!
```
patrickhyw: If it never happens you can `panic!()` here with a description in the panic for why it never happens

optd-datafusion-repr/src/cost/base_cost/stats.rs
```diff
@@ -406,6 +506,7 @@ impl TableStats<Counter<ColumnCombValue>, TDigest<Value>> {
             );
             column_comb_stats.insert(comb, column_stats);
         }
+        println!("First: {:?}, Second: {:?}, Third: {:?}", first, second, now.elapsed());
```
patrickhyw: We should comment this out before merging into main

optd-perftest/src/datafusion_dbms.rs
```diff
@@ -410,7 +414,7 @@ impl DatafusionDBMS {
         &mut self,
         job_kit_config: &JobKitConfig,
     ) -> anyhow::Result<DataFusionBaseTableStats> {
```
patrickhyw, line 551: If you're modifying `get_job_stats()` you should also modify `get_tpch_stats()` in the same way. In the future we can think about creating a trait called `BenchmarkKit` and making a generic function that works with any object of that trait. However, for now unfortunately there's some copied code here.

optd-perftest/src/job.rs
```diff
@@ -137,15 +137,30 @@ impl JobKit {
             .to_string()
     }
 
-    /// Get an iterator through all generated .tbl files of a given config
-    pub fn get_tbl_fpath_iter(&self) -> io::Result<impl Iterator<Item = PathBuf>> {
+    /// Get a vector of all generated .csv files in a given directory path
+    pub fn get_tbl_fpath_vec(&self) -> io::Result<Vec<PathBuf>> {
```
patrickhyw: You should similarly change `get_tbl_fpath_iter()` in `TpchKit` to `get_tbl_fpath_vec()`

patrickhyw (commented):
I left some comments about things to think about.

optd-gungnir/src/stats/misragries.rs
AlSchlo: None

AlSchlo: Data is very skewed, but long term we need a better MG data-structure, and use some count-min

optd-datafusion-repr/src/cost/base_cost/stats.rs
AlSchlo: yes

optd-datafusion-repr/src/cost/base_cost/stats.rs
AlSchlo: done

optd-perftest/src/job.rs
AlSchlo: done

PR #168, opened 2024-04-25
feat: infer equal columns from the query

Gun9niR:
## Background

An investigation of JOB revealed that the reason optd output 1 for all queries was that the selectivities of redundant predicates are all evaluated. A set of predicate contains redundant predicates if some of them can be expressed by others. For example, consider query 1a.

```sql
SELECT mc.note AS production_note,
       t.title AS movie_title,
       t.production_year AS movie_year
FROM company_type AS ct,
     info_type AS it,
     movie_companies AS mc,
     movie_info_idx AS mi_idx,
     title AS t
WHERE ct.kind = 'production companies'
  AND it.info = 'top 250 rank'
  AND mc.note NOT LIKE '%(as Metro-Goldwyn-Mayer Pictures)%'
  AND (mc.note LIKE '%(co-production)%'
       OR mc.note LIKE '%(presents)%')
  AND ct.id = mc.company_type_id
  AND t.id = mc.movie_id
  AND t.id = mi_idx.movie_id
  AND mc.movie_id = mi_idx.movie_id
  AND it.id = mi_idx.info_type_id;
```

In the join plan node with `t` as one of the input tables, the predicates would contain both `t.id = mc.movie_id` and `t.id = mi_idx.movie_id`. But since there's already `mc.movie_id = mi_idx.movie_id`, one of these three predicates is redundant.

## Goal

Given a set of predicates `P` that define the equality of `N` predicates, we want to pick `N - 1` most selective predicates `P'` and remove the rest.

### Identifying Equal Columns

We identify the set of predicates in `GroupColumnRefs` logical property using union find, because this way we can reuse the logic of base table column ref to identify which columns are equal. 

❗ ***However, `BaseTableColumnRef` does not handle table alias, so if `t1` and `t2` are both aliases for `t`, `t1.a = t2.b` and `t1.b = t2.a` will be treated as the same predicate.***

### Computing Selecitivty for `P'`

The difficulty is that we don't get to see all the predicates all at once. Instead, we see those predicates gradually as we move up the plan tree, and the child might have already picked some predicate that is not among `P'`. E.g. in the above example, even if `mc.movie_id = mi_idx.movie_id` (denoted `p3`) is the most selective predicate and thus should have been eliminated, it will be picked by `mc` join `mi_idx` because it's the only predicate it sees.

Therefore, the two predicates involving `t` (denote `t.id = mc.movie_id` as `p1` and `t.id = mi_idx.movie_id` as `p2`) should introduce some "selectivity adjustment factor". More specifically, if we see `p1` first we can just pick it, since `p1` and `p3` are not redundant. When we see `p2`, we need to pick two of the three predicates, but remember `p3` and `p1` are already picked. So in this case instead of just multiplying the total selectivity with `p2`'s, `p2` provides some sort of selectivity adjustment factor which can be computed by `MSP({p1, p2, p3}) / MSP({p1, p3})`, where `MSP` is the multiplied selectivies of the Most Selective Predicates that define the equality of the columns. `MSP` is computed using Kruskal.

This idea can be generalized to more predicates.

## Results

This features significantly improves q-error for the JOB benchmark. Before this PR, the estimated cardinalities for all columns are just `1`.

### DIFF
--- a/optd-datafusion-repr/Cargo.toml
+++ b/optd-datafusion-repr/Cargo.toml
@@ -25,4 +25,4 @@ assert_approx_eq = "1.1.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_with = {version = "3.7.0", features = ["json"]}
 bincode = "1.3.3"
-union-find = { git = "https://github.com/Gun9niR/union-find-rs.git", rev = "31d2ed78c9ca6cd439bbba53b329bdbd7a52da4b" }
+union-find = { git = "https://github.com/Gun9niR/union-find-rs.git", rev = "3ffda352c2c3d9a74daed1546c0e71c97b732bf1" }
--- a/optd-datafusion-repr/Cargo.toml
+++ b/optd-datafusion-repr/Cargo.toml
@@ -25,3 +25,4 @@ assert_approx_eq = "1.1.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_with = {version = "3.7.0", features = ["json"]}
 bincode = "1.3.3"
+union-find = { git = "https://github.com/Gun9niR/union-find-rs.git", rev = "8ea75baced5e15ea0765f899930e21d8caf0dd17" }
--- a/optd-datafusion-repr/Cargo.toml
+++ b/optd-datafusion-repr/Cargo.toml
@@ -25,4 +25,4 @@ assert_approx_eq = "1.1.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_with = {version = "3.7.0", features = ["json"]}
 bincode = "1.3.3"
-union-find = { git = "https://github.com/Gun9niR/union-find-rs.git", rev = "3ffda352c2c3d9a74daed1546c0e71c97b732bf1" }
+union-find = { git = "https://github.com/Gun9niR/union-find-rs.git", rev = "794821514f7daefcbb8d5f38ef04e62fc18b5665" }
--- a/optd-datafusion-repr/Cargo.toml
+++ b/optd-datafusion-repr/Cargo.toml
@@ -25,4 +25,4 @@ assert_approx_eq = "1.1.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_with = {version = "3.7.0", features = ["json"]}
 bincode = "1.3.3"
-union-find = { git = "https://github.com/Gun9niR/union-find-rs.git", rev = "8ea75baced5e15ea0765f899930e21d8caf0dd17" }
+union-find = { git = "https://github.com/Gun9niR/union-find-rs.git", rev = "31d2ed78c9ca6cd439bbba53b329bdbd7a52da4b" }
--- a/optd-datafusion-repr/src/cost/base_cost.rs
+++ b/optd-datafusion-repr/src/cost/base_cost.rs
@@ -317,6 +317,7 @@ mod tests {
 
     pub const TABLE1_NAME: &str = "table1";
     pub const TABLE2_NAME: &str = "table2";
+    pub const TABLE3_NAME: &str = "table3";
 
     // one column is sufficient for all filter selectivity tests
     pub fn create_one_column_cost_model(per_column_stats: TestPerColumnStats) -> TestOptCostModel {
@@ -340,6 +341,41 @@ mod tests {
             tbl2_per_column_stats,
             100,
             100,
+        )
+    }
+
+    /// Two columns is sufficient for all join selectivity tests
+    pub fn create_three_table_cost_model(
+        tbl1_per_column_stats: TestPerColumnStats,
+        tbl2_per_column_stats: TestPerColumnStats,
+        tbl3_per_column_stats: TestPerColumnStats,
+    ) -> TestOptCostModel {
+        OptCostModel::new(
+            vec![
+                (
+                    String::from(TABLE1_NAME),
+                    TableStats::new(
+                        100,
+                        vec![(vec![0], tbl1_per_column_stats)].into_iter().collect(),
+                    ),
+                ),
+                (
+                    String::from(TABLE2_NAME),
+                    TableStats::new(
+                        100,
+                        vec![(vec![0], tbl2_per_column_stats)].into_iter().collect(),
+                    ),
+                ),
+                (
+                    String::from(TABLE3_NAME),
+                    TableStats::new(
+                        100,
+                        vec![(vec![0], tbl3_per_column_stats)].into_iter().collect(),
+                    ),
+                ),
+            ]
+            .into_iter()
+            .collect(),
         )
     }
 
--- a/optd-datafusion-repr/src/cost/base_cost.rs
+++ b/optd-datafusion-repr/src/cost/base_cost.rs
@@ -331,7 +331,7 @@ mod tests {
         )
     }
 
-    /// Two columns is sufficient for all join selectivity tests
+    /// Create a cost model with two columns, one for each table. Each column has 100 values.
     pub fn create_two_table_cost_model(
         tbl1_per_column_stats: TestPerColumnStats,
         tbl2_per_column_stats: TestPerColumnStats,
@@ -344,7 +344,7 @@ mod tests {
         )
     }
 
-    /// Two columns is sufficient for all join selectivity tests
+    /// Create a cost model with three columns, one for each table. Each column has 100 values.
     pub fn create_three_table_cost_model(
         tbl1_per_column_stats: TestPerColumnStats,
         tbl2_per_column_stats: TestPerColumnStats,
--- a/optd-datafusion-repr/src/cost/base_cost.rs
+++ b/optd-datafusion-repr/src/cost/base_cost.rs
@@ -4,7 +4,10 @@ mod join;
 mod limit;
 pub(crate) mod stats;
 
-use crate::{plan_nodes::OptRelNodeTyp, properties::column_ref::ColumnRef};
+use crate::{
+    plan_nodes::OptRelNodeTyp,
+    properties::column_ref::{BaseTableColumnRef, ColumnRef},
+};
 use itertools::Itertools;
 use optd_core::{
     cascades::{CascadesOptimizer, RelNodeContext},
@@ -207,7 +210,7 @@ impl<
         &self,
         col_ref: &ColumnRef,
     ) -> Option<&ColumnCombValueStats<M, D>> {
-        if let ColumnRef::BaseTableColumnRef { table, col_idx } = col_ref {
+        if let ColumnRef::BaseTableColumnRef(BaseTableColumnRef { table, col_idx }) = col_ref {
             self.get_column_comb_stats(table, &[*col_idx])
         } else {
             None
--- a/optd-datafusion-repr/src/cost/base_cost/join.rs
+++ b/optd-datafusion-repr/src/cost/base_cost/join.rs
@@ -347,7 +347,7 @@ impl<
         let mut sorted_predicates = predicates
             .into_iter()
             .map(|p| {
-                let sel = self.get_join_selectivity_from_on_col_ref_pair(
+                let sel: f64 = self.get_join_selectivity_from_on_col_ref_pair(
                     &p.left.clone().into(),
                     &p.right.clone().into(),
                 );
@@ -363,7 +363,7 @@ impl<
                 disjoint_sets.make_set(p.right.clone()).unwrap();
             }
             if !disjoint_sets.same_set(&p.left, &p.right).unwrap() {
-                acc_sel = acc_sel * sel;
+                acc_sel *= sel;
                 num_picked_predicates += 1;
             }
             if num_picked_predicates == num_cols - 1 {
@@ -447,15 +447,18 @@ impl<
                 if let (ColumnRef::BaseTableColumnRef(left), ColumnRef::BaseTableColumnRef(right)) =
                     (left_col_ref, right_col_ref)
                 {
+                    let predicate = EqPredicate::new(left.clone(), right.clone());
                     if past_eq_columns.is_eq(left, right) {
                         return self.get_join_selectivity_from_redundant_predicate(
-                            EqPredicate::new(left.clone(), right.clone()),
+                            predicate,
                             &mut past_eq_columns,
                         );
+                    } else {
+                        past_eq_columns.add_predicate(predicate);
                     }
                 }
 
-                self.get_join_selectivity_from_on_col_ref_pair(&left_col_ref, &right_col_ref)
+                self.get_join_selectivity_from_on_col_ref_pair(left_col_ref, right_col_ref)
             })
             .product()
     }
@@ -468,7 +471,10 @@ mod tests {
     use crate::{
         cost::base_cost::{tests::*, DEFAULT_EQ_SEL},
         plan_nodes::{BinOpType, JoinType, LogOpType, OptRelNodeRef},
-        properties::column_ref::{ColumnRef, GroupColumnRefs},
+        properties::column_ref::{
+            BaseTableColumnRef, ColumnRef, EqBaseTableColumnSets, EqPredicate, GroupColumnRefs,
+            SemanticCorrelation,
+        },
     };
 
     /// A wrapper around get_join_selectivity_from_expr_tree that extracts the table row counts from the cost model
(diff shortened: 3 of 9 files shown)


optd-datafusion-repr/Cargo.toml
```diff
@@ -25,3 +25,4 @@ assert_approx_eq = "1.1.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_with = {version = "3.7.0", features = ["json"]}
 bincode = "1.3.3"
+union-find = { git = "https://github.com/Gun9niR/union-find-rs.git", rev = "3ffda352c2c3d9a74daed1546c0e71c97b732bf1" }
```
patrickhyw: lol nice

optd-datafusion-repr/src/cost/base_cost.rs
```diff
@@ -340,6 +344,41 @@ mod tests {
         )
     }
 
+    /// Two columns is sufficient for all join selectivity tests
```
patrickhyw: should update comment

optd-datafusion-repr/src/cost/base_cost/join.rs
```diff
@@ -114,7 +123,8 @@ impl<
         left_col_cnt: usize,
     ) -> f64 {
         assert!(left_keys.len() == right_keys.len());
-        // I assume that the keys are already in the right order s.t. the ith key of left_keys corresponds with the ith key of right_keys
+        // I assume that the keys are already in the right order
+        // s.t. the ith key of left_keys corresponds with the ith key of right_keys
```
patrickhyw, line 127: Is there a way to assert this assumption?

optd-datafusion-repr/src/cost/base_cost/join.rs
```diff
@@ -303,31 +321,161 @@ impl<
         }
     }
 
+    /// Get the selectivity of one column eq predicate, e.g. colA = colB.
+    fn get_join_selectivity_from_on_col_ref_pair(
```
patrickhyw, line 325: Nice factoring.

optd-datafusion-repr/src/cost/base_cost/join.rs
```diff
@@ -303,31 +321,161 @@ impl<
...
+        let selectivity = ndistincts.map(|ndistinct| 1.0 / ndistinct as f64).reduce(f64::min).expect("reduce() only returns None if the iterator is empty, which is impossible since col_ref_exprs.len() == 2");
+        assert!(
+            !selectivity.is_nan(),
+            "it should be impossible for selectivity to be NaN since n-distinct is never 0"
+        );
+        selectivity
+    }
+
+    /// Given a set of predicates P that define N equal columns, find the selectivity of
+    /// the most selective N - 1 predicates that "touches" all the columns using MST.
```
patrickhyw: You should specify that each predicate is an equality predicate. Also, I know you describe this elsewhere as well, but it would help if you described the undirected weighted graph representation (where columns are nodes, predicates are edges, and selectivities are weights) so that the MST point makes more sense (otherwise people wouldn't know what the "graph") is. Finally, you should mention that the graph should be connected.

optd-datafusion-repr/src/cost/base_cost/join.rs
```diff
@@ -303,31 +321,161 @@ impl<
...
+    fn get_join_selecitivity_from_most_selective_predicates(
+        &self,
+        predicates: Vec<EqPredicate>,
+        num_cols: usize,
+    ) -> f64 {
+        let mut acc_sel = 1.0;
+        let mut num_picked_predicates = 0;
+        let mut disjoint_sets = DisjointSets::new();
+
+        // Sort predicates by selectivity in ascending order.
```
patrickhyw: I think you should mention that you're using Kruskal to improve readability

optd-datafusion-repr/src/cost/base_cost/join.rs
```diff
@@ -303,31 +321,161 @@ impl<
...
+        sorted_predicates.try_for_each(|(p, sel)| {
+            if !disjoint_sets.contains(&p.left) {
+                disjoint_sets.make_set(p.left.clone()).unwrap();
+            }
+            if !disjoint_sets.contains(&p.right) {
+                disjoint_sets.make_set(p.right.clone()).unwrap();
+            }
+            if !disjoint_sets.same_set(&p.left, &p.right).unwrap() {
+                acc_sel *= sel;
+                num_picked_predicates += 1;
```
patrickhyw, line 385: Should there be some sort of operation to "merge" two sets here?

optd-datafusion-repr/src/cost/base_cost/join.rs
```diff
@@ -303,31 +321,161 @@ impl<
...
+            if num_picked_predicates == num_cols - 1 {
+                ControlFlow::Break(())
+            } else {
+                ControlFlow::Continue(())
+            }
+        });
+        assert!(
+            num_picked_predicates == num_cols - 1,
+            "we should have picked N - 1 predicates"
+        );
```
patrickhyw, line 398: It'd be nice to assert that there is a single set in disjoint_sets of size N as well to make sure we didn't make any mistakes

optd-datafusion-repr/src/cost/base_cost/join.rs
```diff
@@ -303,31 +321,161 @@ impl<
...
+        acc_sel
+    }
+
+    /// See `get_join_on_selectivity` for details.
+    fn get_join_selectivity_from_redundant_predicate(
```
patrickhyw: Should it be predicates plural?

optd-datafusion-repr/src/cost/base_cost/join.rs
```diff
@@ -303,31 +321,161 @@ impl<
...
+    /// More specifically, we are checking if the predicate can be expressed with other existing predicates.
+    /// E.g. if we have a predicate like A = B and B = C, we can express A = C is redundant.
+    //
+    /// However, we don't just throw away A = C, because we want to pick the most selective predicates.
+    ///
+    /// More generally, if we have N columns that are equal, and the set of predicates P that
+    /// defines the equality (|P| >= N - 1), we select compute the MST of the graph where the
+    /// columns are nodes and the predicates are edges.
+    ///
+    /// Then the selecitivy of such "redundant" predicate p is the MST of the graph with p (minimum
```
patrickhyw: Typo on selectivity (selecitivy)

optd-datafusion-repr/src/cost/base_cost/join.rs
```diff
@@ -303,31 +321,161 @@ impl<
...
+    /// as in having the smallest product of selectivities) divided by the MST of the graph without p.
```
patrickhyw: Why not move this comment to `get_join_selectivity_from_redundant_predicate()`? I think it makes more sense there

optd-datafusion-repr/src/cost/base_cost/join.rs
```diff
@@ -303,31 +321,161 @@ impl<
...
+
+                if let (ColumnRef::BaseTableColumnRef(left), ColumnRef::BaseTableColumnRef(right)) =
+                    (left_col_ref, right_col_ref)
+                {
+                    let predicate = EqPredicate::new(left.clone(), right.clone());
+                    if past_eq_columns.is_eq(left, right) {
+                        return self.get_join_selectivity_from_redundant_predicate(
+                            predicate,
+                            &mut past_eq_columns,
+                        );
```
patrickhyw, line 499: One possible bug is where a single join operator has two separate predicates that are both in the same connected component. In this case, the selectivity is the selectivity of the MST with _both_ predicates divided by the selectivity of the MST with _neither_ predicate. However, it seems like right now you would compute the selectivity of the MST with _one_ predicate times the selectivity of the MST of the MST with the _other_ predicate divided by the square of the selectivity of the MST with _neither_ predicate, which is not always the same value.

optd-datafusion-repr/src/cost/base_cost/join.rs
```diff
@@ -975,4 +1101,62 @@ mod tests {
             0.02,
         );
     }
+
+    #[test]
+    fn test_inner_redundant_predicate() {
```
patrickhyw, line 1140: Can you describe what specific case you're testing here? I can imagine a few different things to test. For the tests that involve very large graphs, they would involve a lot of joins. However, we'd only need to do an assert on the root join.
1. Adding a predicate which adds another edge to the connected component but keeps the connected component as an MST (for instance, the predicate a = b when we already have b = c).
2. Adding a predicate which adds another edge to a connected component but makes the connected component no longer an MST (for instance, the predicate a = b when we already have a = c and b = c).
  **a.** The new predicate is the least selective.
  **b.** The new predicate is the most selective.
3. Adding a predicate which connects two connected components together where the final graph is not an MST (for instance, the predicate a = b when we already have b = c, c = d, b = d, and a = e, e = f, f = g, a = f).
4. Adding two predicates at once to a connected component where the new component is not an MST (for instance, the predicates a = d and a = e when we already have a = b, b = c, c = d, d = e, a = c
  **a.** The two new predicates are the 1st and 2nd least selective.
  **b.** The two new predicates are the 1st and 2nd most selective.
  **c.** One new predicate is the most selective and the other new predicate is the least selective.

optd-datafusion-repr/src/properties/column_ref.rs
```diff
@@ -1,20 +1,233 @@
...
+    pub fn new(left: BaseTableColumnRef, right: BaseTableColumnRef) -> Self {
+        Self { left, right }
+    }
+}
+
+/// A disjoint set of base table columns with equal values in the same row,
+/// along with the predicates that define the equalities.
+#[derive(Clone, Debug, Default)]
+pub struct EqBaseTableColumnSets {
+    disjoint_eq_col_sets: DisjointSets<BaseTableColumnRef>,
```
patrickhyw, line 102: Do we need to store this? Can't we compute this on-the-fly from the `eq_predicates`? I want to "normalize" the data in the logical properties and not store redundant information.

patrickhyw: Also, if we store `disjoint_eq_col_sets` directly as a logical property, we'd need to write unit tests for the logical property. This feels harder than having a standalone function that computes the disjoint sets from `eq_predicates` which we could unit test on its own.

optd-datafusion-repr/src/properties/column_ref.rs
patrickhyw, line 1: Should we rename the file to `semantic_correlation.rs`?

optd-datafusion-repr/src/properties/column_ref.rs
```diff
@@ -45,71 +263,169 @@ impl PropertyBuilder<OptRelNodeTyp> for ColumnRefPropertyBuilder {
...
+                let exprs = children[1];
+                let column_refs = exprs
+                    .column_refs
+                    .iter()
+                    .map(|p| match p {
+                        ColumnRef::ChildColumnRef { col_idx } => {
+                            children[0].column_refs[*col_idx].clone()
+                        }
+                        ColumnRef::Derived => ColumnRef::Derived,
+                        _ => panic!("projection expr must be Derived or ChildColumnRef"),
```
patrickhyw, line 331: Wait, what if the projection directly contains a `BaseTableColumnRef` just renamed or repositioned? We should add logic (could be in a future PR) that is able to detect semantic correlation even in cases like this.

optd-datafusion-repr/src/properties/column_ref.rs
```diff
@@ -45,71 +263,169 @@ impl PropertyBuilder<OptRelNodeTyp> for ColumnRefPropertyBuilder {
...
+                                }) = &child.output_correlation
+                                {
+                                    eq_column_idx_pairs.extend(pairs.iter());
+                                }
+                            }
+                            Some(SemanticCorrelation {
+                                eq_columns: EqColumns::EqColumnIdxPairs(eq_column_idx_pairs),
+                            })
+                        }
+                        _ => None,
```
patrickhyw, line 315: Shouldn't we still retain the child semantic correlations even if the op is an OR? It's like how we retain it even after a projection

optd-datafusion-repr/src/properties/column_ref.rs
```diff
@@ -119,15 +435,39 @@ impl PropertyBuilder<OptRelNodeTyp> for ColumnRefPropertyBuilder {
...
+                                ) => Some(SemanticCorrelation {
+                                    eq_columns: EqColumns::EqColumnIdxPairs(vec![(
+                                        *l_col_idx, *r_col_idx,
+                                    )]),
+                                }),
+                                _ => None,
+                            }
+                        }
+                    }
+                    _ => None,
```
patrickhyw, line 461: For all the `None` cases, shouldn't we retain the correlation of the children even if the current `BinOp` doesn't add additional correlation?

optd-datafusion-repr/src/properties/column_ref.rs
```diff
@@ -136,3 +476,62 @@ impl PropertyBuilder<OptRelNodeTyp> for ColumnRefPropertyBuilder {
         "column_ref"
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_eq_base_table_column_sets() {
```
patrickhyw, line 485: I didn't quite understand how `EqBaseTableColumnSets` and `EqColumnIdxPairs` worked, but it seemed pretty complex. I think we should add more unit tests (could be in a future PR) testing this functionality (especially with coverage across a broad range of operators).

patrickhyw (changes requested):
Really clean code overall. I left comments about possible bugs in core functionality, more test coverage, readability comments. If the bugs are real we should fix them before merging. The comments should be quick to add so I think you should add them in this PR as well. The test coverage can be done in a future PR since this PR is already so big.

optd-datafusion-repr/src/properties/column_ref.rs
Gun9niR: But the primary purpose of this property is still column ref. I guess we can make a separate folder for this property.

optd-datafusion-repr/src/properties/column_ref.rs
Gun9niR: LogOp does not have a child. It's just a list of expressions. If the operation is `OR`, we are not sure which equality is retained.

optd-datafusion-repr/src/properties/column_ref.rs
Gun9niR: Reposition is the `ChildColumnRef` case, while optd seems to not handle rename at all.

optd-datafusion-repr/src/properties/column_ref.rs
Gun9niR: We are trying to be conservative. For example, what to do when left child is a = b, and right child is a != b?

optd-datafusion-repr/src/properties/column_ref.rs
Gun9niR: It is essentially the same as the `ChildColumnRef` variant in `ColumnRef`. While building the property, we cannot see the entire predicate tree, but only the properties of direct children of a plan node. Look at `derive`'s signature.

```rust
fn derive(
        &self,
        typ: OptRelNodeTyp,
        data: Option<optd_core::rel_node::Value>,
        children: &[&Self::Prop],
    ) -> Self::Prop {
}
```

So say you have a join with predicates #0 = #1, #0 = #2 (grouped together in a `LogOp`), you only get to see the `LopOp`'s property, not the `BinOp`s and `ColumnRefExpr`s. So the point of `EqColumnIdxPairs` is to bubble up the relative indices of `ColumnRefExpr` into the `LogOp`'s property.

optd-datafusion-repr/src/cost/base_cost/join.rs
Gun9niR: Within the cost model, no. It's the *logical join -> hash join* rule's responsibility to ensure this.

optd-datafusion-repr/src/properties/column_ref.rs
Gun9niR: > Do we need to store this? Can't we compute this on-the-fly from the eq_predicates? I want to "normalize" the data in the logical properties and not store redundant information.

You can rebuild the disjoint sets and group it with the predicate set into some new data structure that is only used during cost computation. But since logical property is computed once per group, but compute cost is invoked for every multi-expression in the group, the performance overhead will be significant.

In addition, I don't think maintaining "repetitive" information is problematic if there's a big performance gain. It's a common optimization practice.

Gun9niR: > Also, if we store `disjoint_eq_col_sets` directly as a logical property, we'd need to write unit tests for the logical property. This feels harder than having a standalone function that computes the disjoint sets from `eq_predicates` which we could unit test on its own.

Even if you compute the disjoint sets from `eq_predicates`, you still need to unit test that right? It's the difference between starting from an disjoint sets and starting from a **correct** predicate set and disjoint sets of the children, since `add_predicate` is just add the predicate to the predicate set, and add the columns to disjoint sets. They are basically testing the same thing.

optd-datafusion-repr/src/cost/base_cost/join.rs
Gun9niR: Added comments for this test. Other test cases will be in future PR.

optd-datafusion-repr/src/properties/column_ref.rs
patrickhyw: It's ok, I think we can keep it as `column_ref.rs`

optd-datafusion-repr/src/properties/column_ref.rs
patrickhyw: If we make it a `ChildColumnRef` though, would it still be able to detect correlations? `EqColumns` only stores sets of `BaseTableColumnRef`.

optd-datafusion-repr/src/properties/column_ref.rs
patrickhyw: Why do we need to know the relative indices?

optd-datafusion-repr/src/cost/base_cost/join.rs
Gun9niR: As discuss offline, such scenario will not happen.
image

optd-datafusion-repr/src/properties/column_ref.rs
Gun9niR: Because the build process is bottom up, and `ColumnRefExpr`s are all relative indices. For example a join condition may look like `#0 = #1`. We can't directly see the `ColumnRefExpr`s, so these indices should be kept in the property.

optd-datafusion-repr/src/properties/column_ref.rs
Gun9niR: `ChildColumnRef` is a variant of `ColumnRef`, which has nothing to do with correlation. For correlation detection with relative indices, see `EqColumnIdxPairs`

patrickhyw (approved):
LGTM

PR #173, opened 2024-04-30
[WIP] Physical props

AveryQi115:
- [x] basic framework
- [x] default physical properties builder (without any constraints)
- [x]  test default builder
- [ ] ordering physical properties builder (Sort -> order)
- [ ] testcases
- [ ] code reconstruction

### DIFF
--- a/optd-core/src/cascades.rs
+++ b/optd-core/src/cascades.rs
@@ -5,5 +5,5 @@ mod optimizer;
 mod tasks;
 
 use memo::Memo;
-pub use optimizer::{CascadesOptimizer, GroupId, OptimizerProperties, RelNodeContext};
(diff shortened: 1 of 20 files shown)


jurplel:
Why is it that we needed to incorporate the notion of a `SubGroup` here? I think that is non-standard... 

I am predicting it may be because we are lacking separation of the phases that apply transformation rules and implementation rules?

AveryQi115:
> Why is it that we needed to incorporate the notion of a `SubGroup` here? I think that is non-standard...
> 
> I am predicting it may be because we are lacking separation of the phases that apply transformation rules and implementation rules?

No, it’s not related to separation of transformation rules and impl rules. It’s a standard way to do this, just see calcite and cockroaches code. They both have a notion of logically equivalent and physically equivalent subset in logically equivalent set

PR #175, opened 2024-04-30
feat: cost model handles cast

patrickhyw:
**Summary**: The cost model can now handle `Cast` nodes around `ColumnRef`s and/or `Constant`s

**Demo**:
With this feature, our average q-error on joblight _when using our own optimizer_ is now 5.03 instead of ~4000.
Screenshot 2024-04-30 at 16 37 29

**Details**:
* When using DataFusion's optimizer, the cost model never deals with `Cast` nodes. However, when using the optd optimizer, the cost model does have to deal with `Cast` nodes.
* We should also add a rule in the optimizer to get rid of `Cast`s. However, even after adding that rule, it's still valuable for the cost model to be able to handle `Cast`s.
* The algorithm currently will not work for nested `Cast`s. You'll need to resolve the `Cast`s from bottom-up instead of top-down to do this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -964,8 +964,10 @@ checksum = "7f2c685bad3eb3d45a01354cedb7d5faa66194d1d58ba6e267a8de788f79db38"
 dependencies = [
  "android-tzdata",
  "iana-time-zone",
+ "js-sys",
  "num-traits",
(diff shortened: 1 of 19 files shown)


Gun9niR (approved):
LGTM

PR #178, opened 2024-05-01
fix: ignore null_frac when computing range selectivity

Gun9niR:
Distribution and MCV already accounts for NULLs, so there's no need to multiply range selectivity by `1 - null_frac`.


AlSchlo (approved):
We should adapt the tests to not reduce coverage

Gun9niR:
> We should adapt the tests to not reduce coverage

Coverage should not be affected.

PR #179, opened 2024-05-01
fix: correctly handle `null_frac` in LIKE

Gun9niR:
Since MCV already considers null_frac, we can remove `* (1 - null_frac)`.


patrickhyw (approved):
LGTM

PR #180, opened 2024-05-01
feat: unnesting arbitrary subqueries (likely broken)

jurplel:
Somewhere between a proof of concept and a draft—work still heavily in progress. Will already successfully parse and fully unnest a subset of correlated and uncorrelated subqueries (although I am uncertain about correctness).

**TODO**:
- [ ] Formal testing
- [ ] EXISTS clauses
- [ ] IN clauses
- [ ] ANY/ALL clauses
- [ ] Correctness issue with COUNT(*) (requires adding left outer join to plan)
- [x] Move some/all of this to rewriting stage to support multiple subqueries/ordering operations
- [x] “Sideways information passing” (subplans are duplicated now instead of making a DAG)
    - It seems that a DAG representation is only supported by looking for groups that appear the same. It looks to me that the cloned branches generated by this PR are indeed marked with the same group ID. I marked this bullet point as completed with this in mind.
- [ ] Support more pushdowns (e.g. limit, joins)
- [ ] Optimizations from the paper are all missing (Out of scope?)


skyzh (approved):
rubber stamp approval -- let's break things and fix things

skyzh:
TODO on the tooling side:
* `with_logical` is super confusing, this means "with data fusion logical optimizer", we should either doc it or have a human-readable name
* I think now we need `explain_after_heuristic` so that we can print out a so-called optimized logical plan (subquery un-nesting happens in the heuristic optimizer).
* Probably also need to fix the heuristic optimizer things?
* Also we should consider bump compiler version and upgrade to Rust 2024 edition

skyzh:
also we should consider bump compiler version and upgrade to Rust 2024 edition

PR #186, opened 2024-10-14
2024 `optd` Goals

connortsui20:
This is a first draft of a document detailing the high-level goals for `optd` that we have been talking about over the past few weeks. I've summarized it as 4 main goals, but there might be some higher-level visions that I'm missing.

There are still a lot of TODOs, mainly places where I don't really know what the decisions were or what the concept behind them is. One of the more important TODO places is the database schema layout, I don't fully understand the "left side of the whiteboard" (from our meetings).

There's also a very incomplete design section at the bottom with a few notes on things we talked about in our more recent meetings (I likely don't have time to go through that and make it more detailed any time soon).

My hope is that we can solidify some ideas now in writing so that we can start actually making progress, as well as to begin documenting the decisions we're making so that people can look back on this and understand why we've decided on what have done (really what will will do).

Feel free to comment, though this took a while and I need to focus on other things so I probably won't be able to address them quickly. Please make changes as necessary.

### DIFF
--- a/GOALS.md
+++ b/GOALS.md
@@ -1,0 +1,151 @@
+# The `optd` Query Optimizer
+
+**10/13/2024**
+
+The `optd` optimizer is based on the cost-based Cascades query optimization framework. Many of the terms used in this document will be based on terms and concepts defined in the Microsoft article _Extensible query optimizers in practice_ (specifically in Chapter 2).
+
+The purpose of this document is to record the high-level goals and potential contributions of the `optd` query optimizer project. Most of the ideas detailed in this document were discussed during the first handful of `optd` meetings in Fall 2024.
+
+Note that at the time of writing, `optd` already has a prototype implementation. However, this document purposefully does not reference the current code of the prototype. For the sake of clarify, we will make the assumption that nothing has been implemented, and this project is starting from scratch.
+
+Not all of the goals we describe in this document may be possible. However, by recording all of our goals in one place, we hope to create a unified vision for the `optd` project that will allow us to pave a clearer path.
+
+# Overarching Goals
+
+The main goal of the `optd` project is to build a query optimizer that has the following properties, listed in no particular order of importance:
+
+-   **Standalone**: The `optd` optimizer should be a standalone component. This is in contrast to other query optimizers that are built directly into end-to-end database management systems. The goal of building a standalone optimizer is to allow future DBMS developers to "plug-and-play" `optd` into their own systems, either as an embedded library or as a service.
+-   **Extensible**: Developers that use the `optd` optimizer should be able to customize the behavior of `optd` by adding their own relational operators and rules (among other things). The exact behavior of the execution operators that `optd` is optimizing over should be abstracted away from the internal optimization engine.
+-   **Parallel**: In order to extract the most performance out of modern concurrent hardware (CPUs and storage devices), the query optimization algorithm of `optd` should support efficient parallel search and optimization.
+-   **Persistent**: Queries often repeat, and different query plans often still have similar structure. By persisting state between the optimization of query plans, we can leave a trail of "breadcrumbs" for any of the decisions made for future queries to utilize.
+
+We will explain in depth what each of these goals entail.
+
+## Standalone
+
+The `optd` optimizer should be its own modularized component that DBMS developers can easily "plug in" to their own system. Due to its standalone nature, `optd` should use a standard representation of query plans as the bridge between `optd` and the external execution engine.
+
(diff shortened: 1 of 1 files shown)


jurplel:
I would really prefer not to see a design document dropped into the root repo.

Is it possible to make this into a google doc and share it internally?

This also should be reconciled with the design doc that I have been working on.

GOALS.md
```diff
@@ -0,0 +1,151 @@
...
+
+Contention is still a problem in an asynchronous setting, as there is only so much a runtime can do to alleviate all tasks attempting to access global locks on data structures. Thus, the memo table itself needs to support parallel access and manipulation. At a minimum, the memo table should support finer granularity of locking in the memo table. In the future, it is likely possible to implement the entire memo table in a lock-free manner (copy-on-write, append-only, compare-and-swap semantics, etc.). Using an asynchronous runtime certainly helps, as the cost of "blocking" is not the same as when using traditional synchronous threads.
+
+## Persistent
+
+As far as we are aware, current query optimizers are completely volatile. In other words, once a query plan has been optimized, the system forgets all of the steps it took to reach the final fully-optimized query plan.
```
jurplel, line 76: This isn't true. Query optimizers are not always completely volatile. Query plan caching is pretty common—caching the memo table state is the slightly more unique aspect, although I am not certain that is totally unique (at least at plan granularity)

connortsui20:
> I would really prefer not to see a design document dropped into the root repo.

You can copy and paste it in then. I wanted to have something that was tracked my git and viewable by everyone, but it's not that important.

GOALS.md
connortsui20: Feel free to change it.

GOALS.md
```diff
@@ -0,0 +1,151 @@
...
+
+We will explain in depth what each of these goals entail.
+
+## Standalone
+
+The `optd` optimizer should be its own modularized component that DBMS developers can easily "plug in" to their own system. Due to its standalone nature, `optd` should use a standard representation of query plans as the bridge between `optd` and the external execution engine.
+
+`optd` will use **[Substrait]** as its query plan representation by default. This can potentially be extended to support other representations in the future, but we will use [Substrait] by default.
+
+TODO justification on why we want to use [Substrait]
```
SarveshOO7, line 30: correct me if I am wrong but the only justification we have is that substrait is the only generally used/supported open source SQL IR right now?

GOALS.md
```diff
@@ -0,0 +1,151 @@
...
+
+TODO more information on what integration with [Substrait] should look like
```
SarveshOO7, line 32: Simple, optd should take in a substrait logical plan and return optimized substrait physical plans?

GOALS.md
```diff
@@ -0,0 +1,151 @@
...
+
+Additionally, `optd` will use **[DataFusion]** as its default execution engine as well as use [DataFusion]'s SQL parser and binder.
+
+TODO more information on what integration with [DataFusion] should look like
+
+TODO plans for eventually removing the [DataFusion] dependency in the future?
```
SarveshOO7, line 38: Not clear on what exact dependency you want to remove here. Do you not want to use Datafusion in the future as the primary execution engine for Opt-D? Does it really matter since we are building the optimizer to use substrait? So, anything that uses substrait will always work?

GOALS.md
```diff
@@ -0,0 +1,151 @@
...
+
+The core search algorithm of query optimization itself should not need to know the exact behavior of the operators, relations, or rules it is using for optimization, regardless of if they are provided by `optd` (proprietary) or provided by a third-party developer. This means that the `optd` search engine must rely on dynamic dispatch for the manipulation of external third-party types.
+
+However, this does not mean we must solely switch to an object-oriented model and give up on Rust's powerful type system (abstract data types via `enum`s). Internally, we can use Rust's `enum`s (e.g., via the [`enum_dispatch`] crate) for our proprietary rules and operators that we write ourselves. Then, we can add a `Box<dyn Trait>` variant that allows external developers to additionally add their own types. This would allow compile-time guarantees for the all of the proprietary rules and operators in the standalone library crate, while also allowing for the flexibility of external trait objects. Note that while this might seem like redundant work, it will allow for faster velocity of change for the core components of `optd` while ensuring correctness throughout development.
```
SarveshOO7, line 52: Nit: I think we should try to avoid `Box <dyn Trait> `as much as possible because figuring out what exact struct the Box contains is always a non-trivial operation. @connortsui20 and I have seen this firsthand in the datafusion codebase when building eggstrain.

jurplel:
I think let's move this to Google docs for now.

PR #189, opened 2024-10-26
refactor(sqlplannertest): split directories, allow subdirectory selection, rename with_logical

skyzh:
* with_logical is now with_df_logical (which means that with datafusion logical optimizer)
* apply planner test now supports subdirectories, so that we don't run all tests, that's super slow
* apply planner test now shows time spent on a test case


yliang412:
Could also change to `use_df_logical` here: 

https://github.com/cmu-db/optd/blob/066ae17770c60fdc1647defe312e4682310a6452/datafusion-optd-cli/src/main.rs#L139-L140

yliang412 (approved):

skyzh:
ahh, I enabled auto merge, will fix in the next patch

PR #196, opened 2024-10-28
fix(core): improve dead-loop avoidance in OptimizeInputs

jopereira:
Currently, OptimizeInputsTask avoids a dead-loop with a sequential traversal of a potentially long list of tasks, with a dynamic cast for each of them (see #29). This alternative uses a hashset of previously visited expressions for the same purpose, similarly to explored_group.

Tested only with the TPC-H Q8 demo.


skyzh:
seems that there's no regression in plans

jurplel (approved):

skyzh:
well, there's a regression in self-join, so probably cannot merge for now.

PR #199, opened 2024-10-29
fix(df-repr): join schema inference

skyzh:


jurplel (approved):
Seems right!

skyzh:
Will also fix column ref logical property + add tests in this patch

skyzh:
the bug is not related to schema, it's cost model cannot handle self join. merging the patch for now.

PR #200, opened 2024-10-29
Separate advanced cost model from the previous basic cost model

jurplel:
Tested only briefly, quick and dirty


Cargo.toml
```diff
@@ -7,6 +7,6 @@ members = [
     "optd-sqlplannertest",
     "optd-adaptive-demo",
     "optd-gungnir",
-    "optd-perfbench",
+    "optd-perfbench", "optd-datafusion-repr-adv-cost",
```
skyzh: fmt?

optd-datafusion-repr-adv-cost/src/bin/test_optimize_adv.rs
```diff
@@ -0,0 +1,109 @@
+use std::{collections::HashMap, sync::Arc};
```
skyzh, line 1: make it a unit test? (can be done later)

skyzh (approved):
need to run regression tests, likely a lot of plans have changed

jurplel:
An argument could be made I think to separate the ColumnRef property too, but i'll leave it for now

PR #201, opened 2024-10-29
fix(cost): return full selectivity for self joins

skyzh:
will go after https://github.com/cmu-db/optd/pull/200


optd-datafusion-repr/src/cost/base_cost/join.rs
```diff
@@ -450,6 +450,10 @@ impl<
         predicate: &EqPredicate,
         past_eq_columns: &mut EqBaseTableColumnSets,
     ) -> f64 {
+        if predicate.left == predicate.right {
+            // self-join, TODO: is this correct?
```
jurplel, line 455: Well, it's something for now

jurplel (approved):

PR #203, opened 2024-10-29
refactor(core): rewrite memo table + enable filter-project rules in df-repr

skyzh:
* rewrite the memo table to fully resolve the group merging issues; add test cases; drop a few stale APIs from the memo table
* enable the filter-project set of rules and it won't run out of budget now


jurplel (approved):
Amazing, even some tests

Issue #194, opened 2024-10-27
revive join order enumeration + join reordering bench


skyzh:
Before #192, `explain` in optd by default enumerates all join orders. The functionality was added in order to ensure the current join commute + assoc rule can enumerate all join orders. However, the implementation of join order enumeration is super slow and was opt-in by default when explaining. After removing this functionality, TPC-H test suite is 10x faster on my local setup.

* Find better way to enumerate all join orders (avoid generating all bindings)?
* Have a flag + SQL syntax in the optimizer to enable/disable join order enumeration during explain?
* Test how fast is the cascades optimizer to enumerate all join orders, and look into if we need macro rules to do join reordering.

PR #204, opened 2024-10-30
feat(df-repr): add back join order enumeration

skyzh:
ref https://github.com/cmu-db/optd/issues/194

after the memo table refactor, adding back a more efficient join order enumeration implementation.


skyzh:
(physical join not supported for now, and I don't think it's necessary to support it)

optd-core/src/cascades/optimizer.rs
```diff
@@ -369,6 +369,10 @@ impl<T: RelNodeTyp> CascadesOptimizer<T> {
             .map(|x| x.cost.0[0])
             .unwrap_or(0.0)
     }
+
+    pub fn memo(&self) -> &Memo<T> {
```
jurplel, line 373: Exposing the memo table publicly strikes me as a bit scary. I am worried users of the library might try to manipulate the memo table manually.

I'm hoping access is read-only (looks like it is?)

optd-datafusion-bridge/src/lib.rs
```diff
@@ -258,52 +169,17 @@ impl OptdQueryPlanner {
...
-            //         join_orders.iter().map(|x| x.to_string()).join("\n"),
-            //     ));
-            //     explains.push(StringifiedPlan::new(
-            //         PlanType::OptimizedPhysicalPlan {
-            //             optimizer_name: "optd-all-logical-join-orders".to_string(),
-            //         },
-            //         logical_join_orders.iter().map(|x| x.to_string()).join("\n"),
-            //     ));
-            // }
+            let join_orders = optimizer
```
jurplel, line 172: Is this optional or always calculated?

optd-datafusion-repr/src/memo_ext.rs
```diff
@@ -0,0 +1,186 @@
+//! Memo table extensions
```
jurplel, line 1: definitely much cleaner than in the bridge.

1. I think we'll need to be careful to document it so that it isn't misused by users
2. I think the persistent memo table stuff might actually make this extension system unnecessary, since we can read the database to see which join orders were enumerated.

For now though, definitely an improvement!

jurplel (approved):

optd-core/src/cascades/optimizer.rs
skyzh: yes it's read only 🤪

optd-datafusion-bridge/src/lib.rs
skyzh: always calculated, as before, unless we can find a better way of passing options through datafusion SQL...

optd-datafusion-repr/src/memo_ext.rs
skyzh: the join information is not *directly* persisted in the database? we still need to go through this depth-first search algorithm to reconstruct such info.

jurplel: Right, but then we can do it after the query runs instead of checking the memo table "in vivo"

optd-datafusion-bridge/src/lib.rs
skyzh: that's why I didn't close the issue, this is mentioned in the issue

skyzh:
enable auto-merge...

PR #207, opened 2024-10-31
refactor(repr): empty relation has child, add values node

skyzh:
Empty relation now has a child, this makes it easier to infer logical properties of the node. We don't need to store anything (i.e., schema) in empty relation. Simply wrap the original plan node with empty relation and it's done. (TBD: rename to blackhole node? and remove OneRow variant b/c it's covered by the values node)

Also added values plan node. Currently, this is only used to represent a leaf node that produces nothing.

The difference between values node and empty relation: schema inference. Empty relation passes through all logical properties from the child node (i.e., schema), while values node have its own schema based on the first row of the value (not implemented yet)


optd-sqlplannertest/tests/subqueries/subquery_unnesting.planner.sql
```diff
@@ -64,13 +64,9 @@ PhysicalProjection { exprs: [ #0, #1 ] }
...
             ├── groups: [ #1 ]
-            └── PhysicalFilter
-                ├── cond:Eq
-                │   ├── #1
-                │   └── #0
-                └── PhysicalNestedLoopJoin { join_type: Inner, cond: true }
-                    ├── PhysicalAgg { aggrs: [], groups: [ #0 ] }
-                    │   └── PhysicalScan { table: t1 }
-                    └── PhysicalScan { table: t2 }
+            └── PhysicalHashJoin { join_type: Inner, left_keys: [ #0 ], right_keys: [ #0 ] }
```
skyzh: mysterious plan improvement...?

skyzh: resolved in https://github.com/cmu-db/optd/pull/208

jurplel:
> remove OneRow variant b/c it's covered by the values node

Are we using it for anything now? Maybe we can just remove it.

> rename to blackhole node
EmptyRelation matches datafusion but i think blackhole is a more meaningful name


I am still not sure about this change, though. I wish that we could find a way to represent the blackhole/emptyrelation as a node with no children

jurplel (commented):
> remove OneRow variant b/c it's covered by the values node

Are we using it for anything now? Maybe we can just remove it.

> rename to blackhole node

EmptyRelation matches datafusion but i think blackhole is a more meaningful name


I am still not sure about this change, though. I wish that we could find a way to represent the blackhole/emptyrelation as a node with no children

optd-datafusion-repr/src/properties/column_ref.rs
```diff
@@ -294,17 +293,6 @@ impl PropertyBuilder<OptRelNodeTyp> for ColumnRefPropertyBuilder {
...
-                let data = data.unwrap().as_slice();
-                let empty_relation_data: EmptyRelationData =
-                    bincode::deserialize(data.as_ref()).unwrap();
-                let schema = empty_relation_data.schema;
-                let column_cnt = schema.fields.len();
-                let column_refs = (0..column_cnt)
-                    .map(|i| ColumnRef::base_table_column_ref(DEFAULT_NAME.to_string(), i))
-                    .collect();
-                GroupColumnRefs::new(column_refs, None)
-            }
```
skyzh, line 307: the key change here is we did a passthrough of logical properties instead of storing them inside the plan node

skyzh:
let's hold this patch and discuss it later, I'm also unsure about how to represent such empty thing either as a property or as a plan node

Issue #107, opened 2024-03-13
Pruning Support


yliang412:
Resources:
- [RelMdLowerBoundCost](https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/rel/metadata/RelMdLowerBoundCost.java)
- [TopDownRuleDriver.java](https://github.com/apache/calcite/blob/b0034c9ed885644e22edc92e8ab939128ee6b615/core/src/main/java/org/apache/calcite/plan/volcano/TopDownRuleDriver.java#L618)
- Related: PR https://github.com/apache/calcite/pull/1991

PR #210, opened 2024-11-02
feat(core): support upper bound cost-based pruning

### LINKED ISSUE #9: cost model - compute lower / upper bound
currently, we do not compute upper bound / lower bound for the pruning process. we simply explore the full logical plan space and compute a winner for the physical plan. the pruning logic is already there, and we just need to provide a lower / upper bound of groups to the system.

skyzh:
actually with pruning implemented, I don't see a lot of improvements in plan generation time... 🤣

ref https://github.com/cmu-db/optd/issues/107, we still need lower bound cost
close https://github.com/cmu-db/optd/issues/9

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3886,7 +3886,7 @@ dependencies = [
 [[package]]
 name = "sqlplannertest"
 version = "0.1.0"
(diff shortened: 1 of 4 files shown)


jurplel:
I think this gets overwritten by the task graph rewrite (and the logic looks different according to the book also). maybe the book logic will have better performance

no reason not to merge though, i suppose

jurplel (approved):

PR #211, opened 2024-11-02
refactor(core): abstract memo table into a trait

skyzh:
The current memo table implementation is refactored into a trait, and the implementation is called `NaiveMemo`.


optd-core/src/cascades/memo.rs
```diff
@@ -98,13 +98,147 @@ pub struct GroupInfo {
...
+                panic!("group {group_id} has {len} expressions")
+            } else {
+                None
+            }
+        }
+    }
+}
+
+/// A naive, simple, and unoptimized memo table implementation.
+pub struct NaiveMemo<T: RelNodeTyp> {
```
jurplel, line 241: We should move this impl to another file at some point

jurplel (approved):

PR #212, opened 2024-11-02
fix(df-repr): depjoin elimination rule

skyzh:
This is not the most efficient way to implement this, but at least it works. Reimplemented the eliminate depjoin rule, so that it can remove depjoin and insert join when there are no correlated columns. This is done by recursively inspecting whether `ExternCol` exists on the right side of the plan tree. This also means that this eliminate depjoin rule must be used as a heuristics rule.

Added a new subquery regression test and enabled TPC-H 1-5 to use the optd logical optimizer.


skyzh:
If we want to implement it as a cascades rule, the most obvious way I can think of is to add a `HasCorrelatedColumn` logical property. Probably we also need to record what correlated columns are there in this property, so that it can handle nested depjoin correctly.

optd-datafusion-repr/src/rules/subquery/depjoin_pushdown.rs
```diff
@@ -418,8 +418,10 @@ fn apply_dep_join_past_agg(
     vec![new_agg.into_rel_node().as_ref().clone()]
 }
 
+// Heuristics-only rule. If we don't have references to the external columns on the right side,
```
jurplel, line 421: I wonder if we can detect if a rule can only be done heuristically automatically? Can't think of any easy way.

Maybe we can set a flag when we define the rule with the macro (to prevent misuse)

optd-datafusion-repr/src/rules/subquery/depjoin_pushdown.rs
```diff
@@ -428,31 +430,43 @@ define_rule!(
...
-    DepJoinEliminateAtScanPicks {
+    DepJoinEliminatePicks {
         left,
         right,
         cond,
         extern_cols: _,
-    }: DepJoinEliminateAtScanPicks,
+    }: DepJoinEliminatePicks,
 ) -> Vec<RelNode<OptRelNodeTyp>> {
     // TODO: Is there ever a situation we need to detect that we can convert earlier?
```
jurplel: I think you fixed this TODO

jurplel (approved):

jurplel:
great work!

PR #219, opened 2024-11-05
refactor(core): add predicate into relnode

skyzh:
Extracted from https://github.com/cmu-db/optd/pull/193, this patch adds predicate to the relnode and refactors the memo table to store the predicates. Minimum changes are done to the df-repr to ensure it still works.

Follow-ups:
* Rename RelNode -> PlanNode
* Rename RelNodeTyp -> PlanNodeTyp
* Refactor RelNode to store RelNodeOrGroup
* Remove `data` from RelNode

### DIFF
--- a/optd-core/src/cascades/memo.rs
+++ b/optd-core/src/cascades/memo.rs
@@ -11,10 +11,10 @@ use tracing::trace;
 use crate::{
     cost::{Cost, Statistics},
     property::PropertyBuilderAny,
-    rel_node::{RelNode, RelNodeRef, RelNodeTyp, Value},
+    rel_node::{ArcPredNode, RelNode, RelNodeRef, RelNodeTyp, Value},
 };
 
-use super::optimizer::{ExprId, GroupId};
+use super::optimizer::{ExprId, GroupId, PredId};
 
 pub type RelMemoNodeRef<T> = Arc<RelMemoNode<T>>;
 
@@ -24,6 +24,7 @@ pub struct RelMemoNode<T: RelNodeTyp> {
     pub typ: T,
     pub children: Vec<GroupId>,
     pub data: Option<Value>,
+    pub predicates: Vec<PredId>,
 }
 
 impl<T: RelNodeTyp> RelMemoNode<T> {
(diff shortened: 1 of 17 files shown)


optd-core/src/cascades/memo.rs
```diff
@@ -247,6 +263,10 @@ pub struct NaiveMemo<T: RelNodeTyp> {
     groups: HashMap<GroupId, Group>,
     expr_id_to_expr_node: HashMap<ExprId, RelMemoNodeRef<T>>,
 
+    // Predicate stuff. We don't find logically equivalent predicates. Duplicate predicates
+    // will have different IDs.
+    pred_id_to_pred_node: HashMap<PredId, ArcPredNode<T>>,
+
```
jurplel, line 269: What happened to `pred_node_to_pred_id`? With that table we could've found equivalent predicates i think

jurplel:
Another follow-up: Remove `data` field

optd-core/src/cascades/memo.rs
```diff
@@ -368,6 +399,13 @@ impl<T: RelNodeTyp> NaiveMemo<T> {
         ExprId(id)
     }
 
+    /// Get the next pred id. Group id and expr id shares the same counter, so as to make it easier to debug...
```
jurplel, line 402: nice idea!

jurplel (approved):
Seems like a solid start

optd-core/src/cascades/memo.rs
skyzh: We don't need to do that because finding equivalents is only needed for cascades, but we don't plan to run cost-based search on expressions?

skyzh: Furthermore, b/c we always do full match on predicates, it doesn't make sense to generate bindings and do equivalent matching for predicates

skyzh: ... if there's a predicate rule, it will always create a new predicate ID

jurplel: Fair point. I was thinking in terms of the "predicate group" idea from our discussions, I suppose. This should work.

PR #221, opened 2024-11-05
refactor(core,repr): expose group type in core

skyzh:
rule WIP


skyzh:
took over by https://github.com/cmu-db/optd/pull/223, probably I'll do some of this patch's refactor later

PR #223, opened 2024-11-06
refactor: RelNode/RexNode (PlanNode/PredNode) separation + more refactors

skyzh:
please rebase + merge


skyzh:
... and need to rewrite the history to include coauthors

skyzh:
seems like there are some regressions on TPC-H where filters are not pushed down, needs investigation

optd-core/src/rules.rs
```diff
@@ -58,9 +55,11 @@ impl<T: RelNodeTyp, O: Optimizer<T>> RuleWrapper<T, O> {
     }
 }
 
-pub trait Rule<T: RelNodeTyp, O: Optimizer<T>>: 'static + Send + Sync {
+// TODO: docs, possible renames.
+// TODO: Why do we have all of these match types? Seems like possible overkill.
+pub trait Rule<T: NodeType, O: Optimizer<T>>: 'static + Send + Sync {
     fn matcher(&self) -> &RuleMatcher<T>;
-    fn apply(&self, optimizer: &O, input: HashMap<usize, RelNode<T>>) -> Vec<RelNode<T>>;
+    fn apply(&self, optimizer: &O, binding: ArcPlanNode<T>) -> Vec<PlanNodeOrGroup<T>>;
```
jurplel, line 62: Simpler interface seems like a good idea!

optd-datafusion-repr/src/plan_nodes.rs
```diff
@@ -204,321 +137,75 @@ pub trait OptRelNode: 'static + Clone {
...
-            .get(&(self.0.as_ref() as *const _ as usize))
+            .get(&(self.clone().into_plan_node().as_ref() as *const _ as usize))
             .unwrap()
     }
 }
 
-impl OptRelNode for PlanNode {
-    fn into_rel_node(self) -> OptRelNodeRef {
-        self.0
+impl DfReprPlanNode for ArcDfPlanNode {
```
jurplel, line 147: Instead of implementing the whole trait on arcdfplannode, i thought it was better to use external dispatch functions for get_meta and dispatch_explain.

I think the tradeoff is that it is a little bit easier to write `.explain()` on an arcdfplannode rather than using an external function, but on the other hand we have a dummy into_plan_node and from_plan_node that don't really make sense and I think make the interface a bit more confusing (i.e. to the user i think it becomes a lot more confusing when they should use into_plan_node and from_plan_node than with the other approach)

Food for thought, doesn't have to be part of this PR

optd-datafusion-repr/src/plan_nodes/join.rs
```diff
@@ -24,42 +24,42 @@ impl Display for JoinType {
...
+    PhysicalHashJoin : DfPlanNode,
     PhysicalHashJoin, [
-        { 0, left: PlanNode },
-        { 1, right: PlanNode }
+        { 0, left: ArcDfPlanNode },
+        { 1, right: ArcDfPlanNode }
     ], [
-        { 2, left_keys: ExprList },
-        { 3, right_keys: ExprList }
+        { 0, left_keys: ListPred },
```
jurplel, line 61: I wonder if there is any reason this is `ListPred` instead of `ArcDfPredNode`.

One thing that comes to mind is that the others have `Arc<DfPredNode>` while this one is actually not behind an `Arc`. Not sure if that matters

optd-datafusion-repr/src/plan_nodes/macros.rs
```diff
@@ -47,44 +46,62 @@ macro_rules! define_plan_node {
...
+                        ],
+                        predicates: vec![
+                            $($attr_name.into_pred_node(),)*
+                        ],
+                    }
+                    .into(),
+                )
+            }
+
+            pub fn new_unchecked(
```
jurplel, line 67: What is `new_unchecked`? I don't see what the difference is with `new` on a quick glance

optd-datafusion-repr/src/plan_nodes/predicates.rs
```diff
@@ -0,0 +1,93 @@
...
+pub use like_pred::LikePred;
+pub use list_pred::ListPred;
+pub use log_op_pred::{LogOpPred, LogOpType};
+use optd_core::nodes::PredNode;
+pub use sort_order_pred::{SortOrderPred, SortOrderType};
+pub use un_op_pred::{UnOpPred, UnOpType};
+
+use super::DfReprPredNode;
+
+pub trait PredExt {
```
jurplel, line 37: Nice

optd-sqlplannertest/tests/basic/cross_product.planner.sql
```diff
@@ -17,7 +17,7 @@ LogicalProjection { exprs: [ #0, #1 ] }
 └── LogicalJoin { join_type: Cross, cond: true }
     ├── LogicalScan { table: t1 }
     └── LogicalScan { table: t2 }
-PhysicalNestedLoopJoin { join_type: Cross, cond: true }
+PhysicalNestedLoopJoin { join_type: Inner, cond: true }
```
jurplel: Is this a correctness problem? I would think cross and inner join were different things

optd-sqlplannertest/tests/tpch/tpch-11-15.planner.sql
```diff
@@ -198,36 +198,36 @@ PhysicalSort
...
-    │           ├── PhysicalHashJoin { join_type: Inner, left_keys: [ #0 ], right_keys: [ #3 ] }
-    │           │   ├── PhysicalFilter
-    │           │   │   ├── cond:Eq
-    │           │   │   │   ├── #1
-    │           │   │   │   └── "CHINA"
-    │           │   │   └── PhysicalScan { table: nation }
-    │           │   └── PhysicalScan { table: supplier }
-    │           └── PhysicalScan { table: partsupp }
+    │       └── PhysicalHashJoin { join_type: Inner, left_keys: [ #0 ], right_keys: [ #3 ] }
+    │           ├── PhysicalFilter
```
jurplel: the filter pushdown issue seems to happen on joins in particular

optd-datafusion-repr/src/rules/filter_pushdown.rs
```diff
@@ -281,32 +250,32 @@ fn filter_join_transpose(
         JoinType::Inner => {
             let old_cond = join_cond;
             let new_conds = merge_conds(and_expr_list_to_expr(join_conds), old_cond);
-            LogicalJoin::new(new_left, new_right, new_conds, JoinType::Inner)
+            LogicalJoin::new_unchecked(new_left, new_right, new_conds, JoinType::Inner)
```
jurplel, line 253: Why are we using `new_unchecked` instead of `new`?

optd-datafusion-repr/src/plan_nodes/sort.rs
```diff
@@ -1,32 +1,31 @@
...
 // 3. actual expr is stored as a child of this node
 define_plan_node!(
-    LogicalSort : PlanNode,
+    LogicalSort : DfPlanNode,
     Sort, [
-        { 0, child: PlanNode }
+        { 0, child: ArcDfPlanNode }
     ], [
-        { 1, exprs: ExprList }
+        { 0, exprs: ListPred }
```
jurplel, line 17: Same concern with listpred over arcdfprednode

optd-datafusion-repr/src/plan_nodes/projection.rs
```diff
@@ -1,28 +1,27 @@
...
 
 define_plan_node!(
-    PhysicalProjection : PlanNode,
+    PhysicalProjection : DfPlanNode,
     PhysicalProjection, [
-        { 0, child: PlanNode }
+        { 0, child: ArcDfPlanNode }
     ], [
-        { 1, exprs: ExprList }
+        { 0, exprs: ListPred }
```
jurplel, line 25: Same concern with listpred over arcdfprednode

skyzh:
new_unchecked: the plan node parameters could be a plan group, which means that we don't check the type of the plan node. predicates will still be checked against the type system.

> I would think cross and inner join were different things

inner/cross joins with true predicate are the same

> One thing that comes to mind is that the others have Arc<DfPredNode> while this one is actually not behind an Arc. Not sure if that matters

This is. The definition is ListPred(ArcDfPredNode), so it gives us a stronger typing contract.

skyzh:
> Instead of implementing the whole trait on arcdfplannode, i thought it was better to use external dispatch functions for get_meta and dispatch_explain.

Yes, we should do this (i.e., I already have PredExt).

jurplel:
> > Instead of implementing the whole trait on arcdfplannode, i thought it was better to use external dispatch functions for get_meta and dispatch_explain.
> 
> Yes, we should do this (i.e., I already have PredExt).

Feel free to refer to my branch, I didn't even use an extension trait but if that works that's good too

https://github.com/cmu-db/optd/blob/bowad/core-hacking-rexnode/optd-datafusion-repr/src/plan_nodes.rs

optd-datafusion-repr/src/plan_nodes.rs
skyzh: decided to do it in another PR

PR #230, opened 2024-11-09
feat(core): support physical properties in heuristics

skyzh:
* add physical property builder
* support building and passing through physical properties in heuristics (we can change the behavior later, i.e., choose to not passthrough but simply add new nodes to satisfy)
* add test cases for heuristics + physical properties

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1605,6 +1605,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "diff"
+version = "0.1.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "56254986775e3233ffa9c4d7d3faaf6d36a2c09d30b20687e9f88bc8bafc16c8"
+
+[[package]]
 name = "difflib"
 version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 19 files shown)


yliang412 (approved):
Overall LGTM! 

From what I understand, we are enforcing one physical prop at a time. Intuitively, it feels like we should make the plan satisfy all constraints at the same time. One idea I have is that we probably could define an enforcer precedence (if we could come up with one) that ensures properties that are already enforced should be kept when enforcing the next one. (The comment here does not block this PR from merging).

skyzh:
Yes -- see the TODO, it is possible that enforcing one property would invalidate the other one. Either we need to do checks, or we should find a better way of integrating this.

jurplel (commented):
As mentioned, I think this is a really great start and I'm happy to see something entering the codebase for us to experiment with, but I am a bit concerned that the design seems non-standard in that it does not support enforcer rules. 

I also would like to see documentation for some of the design decisions, (e..g the two property builder traits) and possibly more documentation on the methods and their intended usage as well, since I find it a bit difficult to understand at a glance (although perhaps not much can be done about that).

skyzh:
> and their intended usage as well

I don't know how they will / they can be used in cascades either, currently I just put them in heuristics.

> I am a bit concerned that the design seems non-standard in that it does not support enforcer rules

I think it's fine to keep the design for now -- we can add enforcer rules in the future (and remove the enforce function), but the derive + passthrough functions are definitely required.

> two property builder traits

basically it's erasing the generics and I can add more docs in the next patch... or we can sit down and talk about what's confusing.

PR #232, opened 2024-11-12
feat(sqllogictest): add initial TPC-H Q1-6 tests

skyzh:
We don't really know if the generated plan is correct or not -- what we all know now is whether we can generate a plan. Therefore, it's important to also evaluate if the result is correct. In this patch, we add sqllogictest, the test infra to ensure the execution result is correct.

In `sqllogictest.rs`, you can replace the following statement to planning + execution solely with datafusion. This yields source-of-truth result.

```
DatafusionDBMS::new_no_optd().await?
```

Now we have TPC-H Q1-6. Note that Q2 doesn't work b/c there are too many NLJs and the datafusion built-in statistics would throw multiply overflow error.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -161,7 +161,7 @@ dependencies = [
  "snap",
  "strum 0.25.0",
  "strum_macros 0.25.3",
- "thiserror",
+ "thiserror 1.0.68",
  "typed-builder",
  "uuid",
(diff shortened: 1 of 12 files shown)


yliang412 (approved):

jurplel (approved):
amazing!

PR #234, opened 2024-11-12
refactor(core): Standardized task graph

jurplel:
A cycle issue has cropped up from this change that is extremely prevalent. Will require looking into further.


datafusion-optd-cli/src/main.rs
```diff
@@ -156,7 +156,7 @@ pub async fn main() -> Result<()> {
     let args = Args::parse();
 
     tracing_subscriber::fmt()
-        .with_max_level(tracing::Level::INFO)
```
jurplel, line 159: TODO: revert this

PR #236, opened 2024-11-13
refactor(core): drop predicate binding interfaces

skyzh:
They are not used any more after the predicate refactor.


jurplel (approved):
Thank you Chi!

PR #237, opened 2024-11-13
feature: dump and visualize the memo as a graph

jopereira:
Dump the state of the Cascades memo to GraphViz .dot files in the current working directory every time a rule is applied.  This is enabled by setting log target optd-memoviz to Trace level. For example, with env_logger, this means setting RUST_LOG=optd-memoviz=trace.

Individual traces can be rendered to a single animated HTML file with ./dev_scripts/memoviz.sh.

Currently, this collects only groups with at least one node that is_logical() to simplify the final result.

### DIFF
--- a/optd-core/src/cascades/optimizer.rs
+++ b/optd-core/src/cascades/optimizer.rs
@@ -5,6 +5,8 @@
 
 use std::collections::{BTreeSet, HashMap, HashSet, VecDeque};
 use std::fmt::Display;
+use std::hash::{DefaultHasher, Hasher};
+use std::io::Write;
 use std::sync::Arc;
 
 use anyhow::Result;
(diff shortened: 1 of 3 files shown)


skyzh:
Generally LGTM. Quick comments:
* I'd like to make this as an optimizer option instead of controlled by log commands. I can push to this pull request branch later.
  * And I need to think about how to integrate with the rest of the system... Should we have a new bin target that takes cmdline input of SQL, or should we have a session variable in the datafusion CLI to control it?
* Dump_dot can take `&mut impl std::io::Write` so that we don't need to wrap it with boxes.
* I feel it would be better to suffix the file with something like a monotonically increasing `step_id` instead of the time?
* With the new refactor, we now store predicates separately instead of in the plan node structure, so we should include it in the graphviz.

jopereira:
Thank you for the feedback. I'm working on it.

jopereira:
I pushed changes for the first three items.

Regarding the fourth, as predicate children are now always materialized, it is harder to draw the actual graph of the memo with aliasing. One possibility is to fully draw each predicate tree wherever it is used.

PR #242, opened 2024-11-16
serialize variant tag

yliang412:
Serialize NodeType and PredType into variant tag (u16/i16).

- Added `strum` for helping with repr conversion.
- Arguably a lot of things can be simplified using a declarative macro.
- Added unit test `test_df_node_type_to_tag_e2e` and `test_df_pred_type_to_tag_e2e`.

### DIFF
--- a/optd-core/src/nodes.rs
+++ b/optd-core/src/nodes.rs
@@ -217,7 +217,10 @@ impl Value {
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
-pub struct VariantTag(pub u16);
+pub struct SerializedNodeTag(pub u16);
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct SerializedPredTag(pub u16);
 
 pub trait NodeType:
     PartialEq
@@ -229,8 +232,8 @@ pub trait NodeType:
     + Debug
     + Send
     + Sync
-    + TryFrom<VariantTag>
-    + Into<VariantTag>
+    + TryFrom<SerializedNodeTag>
+    + Into<SerializedNodeTag>
 {
     type PredType: PartialEq
         + Eq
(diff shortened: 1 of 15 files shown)


optd-datafusion-repr/src/plan_nodes/join.rs
```diff
@@ -11,8 +11,12 @@ use serde::{Deserialize, Serialize};
 use super::macros::define_plan_node;
 use super::{ArcDfPlanNode, ArcDfPredNode, DfNodeType, DfPlanNode, DfReprPlanNode, ListPred};
 
-#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[derive(
+    Debug, Copy, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, Default, strum::FromRepr,
+)]
+#[repr(u8)]
 pub enum JoinType {
+    #[default]
```
yliang412: Despite that `strum::FromRepr` is super useful, I don't like the idea to impl Default for these enums. I'm ok with hand rolling this.

optd-core/src/nodes.rs
```diff
@@ -216,10 +216,33 @@ impl Value {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct VariantTag(pub u16);
```
jurplel: Maybe we should name this `SerializedNodeType` or something

yliang412: 18de24dc9520e69350608429613ba9eec09f0990: Renamed to `SerializedNodeTag` and `SerializedPredTag`.

optd-datafusion-repr/src/plan_nodes/join.rs
yliang412: abandoned using default, instead do a flat enum conversion.

jurplel (approved):
great!

PR #243, opened 2024-11-17
refactor(core): rm option around cost models

skyzh:
`Option` was introduced during a transition time where we thought cost model can compute the cost solely based on the children cost but it turned out that we need the optimizer for a few derived logical properties. We can drop them now.


jurplel (approved):
lgtm

PR #245, opened 2024-11-23
fix(bridge): handle information schema

skyzh:
The information schema is not in datafusion's catalog. It is handled differently as in the upstream source code:

https://github.com/apache/datafusion/blob/c0ca4b4e449e07c3bcd6f3593fa31dd31ed5e0c5/datafusion/core/src/execution/session_state.rs#L301-L311

Therefore, we added this special case in the df-bridge.

### DIFF
--- a/optd-datafusion-bridge/src/lib.rs
+++ b/optd-datafusion-bridge/src/lib.rs
@@ -14,6 +14,7 @@ use std::sync::{Arc, Mutex};
 
 use async_trait::async_trait;
 use datafusion::arrow::datatypes::DataType;
+use datafusion::catalog::information_schema::InformationSchemaProvider;
 use datafusion::catalog::CatalogList;
 use datafusion::error::Result;
 use datafusion::execution::context::{QueryPlanner, SessionState};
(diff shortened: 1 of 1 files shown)


optd-datafusion-bridge/src/lib.rs
```diff
@@ -60,8 +61,24 @@ impl DatafusionCatalog {
...
+            // available.
+            Arc::new(InformationSchemaProvider::new(Arc::clone(&self.catalog)))
+        } else if let Some(schema) = catalog.schema(schema) {
+            schema
+        } else {
+            panic!("schema not found in datafusion catalog: {}", schema)
+        };
+        let Some(table) = futures_lite::future::block_on(schema.table(table.as_ref())) else {
+            panic!("table not found in datafusion catalog: {}", name);
+        };
```
yliang412, line 81: I feel using datafusion's resolve method makes it cleaner:
https://github.com/cmu-db/optd/blob/d5b0239d22cc4416635a099830cd82da1491b7b8/optd-datafusion-bridge/src/lib.rs#L174-L181

I had it while drafting https://github.com/cmu-db/optd/pull/233

PR #251, opened 2024-11-27
feat(ng-kernel): add the persistent kernel

skyzh:
Thinking about the new start point of optd, let's build something from a MVP instead of getting everything working at the first place.

We have naive + persistent memo table. The memo table doesn't store cost and properties for now. It finds the duplicates. It's async.

On the persistent side, we use sqlx to run the queries in in-memory sqlite to eliminate external dependencies when running tests. The persistent memo table currently supports dedup predicates.

### DIFF
--- a/optd-ng-kernel/Cargo.toml
+++ b/optd-ng-kernel/Cargo.toml
@@ -1,0 +1,26 @@
+[package]
+name = "optd-ng-kernel"
+version.workspace = true
+edition.workspace = true
+homepage.workspace = true
+keywords.workspace = true
+license.workspace = true
+repository.workspace = true
+
+[dependencies]
+anyhow = "1"
+async-trait = "0.1"
+arrow-schema = "47.0.0"
+tracing = "0.1"
+ordered-float = "4"
+itertools = "0.13"
+serde = { version = "1.0", features = ["derive", "rc"] }
+chrono = "0.4"
+sqlx = { version = "0.8", features = [
+    "runtime-tokio",
+    "sqlite",
+] } # TODO: strip the features, move to another crate
+serde_json = { version = "1" } # TODO: move to another crate
+
+[dev-dependencies]
+tokio = { version = "1", features = ["full"] }
--- a/optd-ng-kernel/Cargo.toml
+++ b/optd-ng-kernel/Cargo.toml
@@ -9,6 +9,7 @@ repository.workspace = true
 
 [dependencies]
 anyhow = "1"
+async-recursion = "1"
 async-trait = "0.1"
 arrow-schema = "47.0.0"
 tracing = "0.1"
--- a/optd-ng-kernel/src/cascades.rs
+++ b/optd-ng-kernel/src/cascades.rs
@@ -1,0 +1,9 @@
+// Copyright (c) 2023-2024 CMU Database Group
+//
+// Use of this source code is governed by an MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+pub mod memo;
+pub mod naive_memo;
+pub mod optimizer;
+pub mod persistent_memo;
--- a/optd-ng-kernel/src/cascades/optimizer.rs
+++ b/optd-ng-kernel/src/cascades/optimizer.rs
@@ -1,0 +1,33 @@
+// Copyright (c) 2023-2024 CMU Database Group
+//
+// Use of this source code is governed by an MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+use std::fmt::Display;
+
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug, Default, Hash)]
+pub struct GroupId(pub(super) usize);
+
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug, Default, Hash)]
+pub struct ExprId(pub usize);
+
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug, Default, Hash)]
+pub struct PredId(pub usize);
+
+impl Display for GroupId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "!{}", self.0)
+    }
+}
+
+impl Display for ExprId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+impl Display for PredId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "P{}", self.0)
+    }
+}
--- a/optd-ng-kernel/src/nodes.rs
+++ b/optd-ng-kernel/src/nodes.rs
@@ -224,7 +224,12 @@ pub trait NodeType:
 
 pub trait PersistentNodeType: NodeType {
     fn serialize_pred(pred: &ArcPredNode<Self>) -> serde_json::Value;
+
     fn deserialize_pred(data: serde_json::Value) -> ArcPredNode<Self>;
+
+    fn serialize_plan_tag(tag: Self) -> serde_json::Value;
+
+    fn deserialize_plan_tag(data: serde_json::Value) -> Self;
 }
 
 /// A pointer to a plan node
--- a/optd-ng-kernel/src/nodes.rs
+++ b/optd-ng-kernel/src/nodes.rs
@@ -1,0 +1,365 @@
+// Copyright (c) 2023-2024 CMU Database Group
+//
+// Use of this source code is governed by an MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+//! The RelNode is the basic data structure of the optimizer. It is dynamically typed and is
+//! the internal representation of the plan nodes.
+
+use std::fmt::{Debug, Display};
+use std::hash::Hash;
+use std::sync::Arc;
+
+use arrow_schema::DataType;
+use chrono::NaiveDate;
+use ordered_float::OrderedFloat;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+use crate::cascades::optimizer::GroupId;
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct SerializableOrderedF64(pub OrderedFloat<f64>);
+
+impl Serialize for SerializableOrderedF64 {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        // Directly serialize the inner f64 value of the OrderedFloat
+        self.0 .0.serialize(serializer)
+    }
+}
+
+impl<'de> Deserialize<'de> for SerializableOrderedF64 {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        // Deserialize an f64 and wrap it in an OrderedFloat
+        let float = f64::deserialize(deserializer)?;
+        Ok(SerializableOrderedF64(OrderedFloat(float)))
+    }
+}
+
+// TODO: why not use arrow types here? Do we really need to define our own Value type?
+// Shouldn't we at least remove this from the core/engine?
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize, PartialOrd, Ord)]
+pub enum Value {
+    UInt8(u8),
+    UInt16(u16),
+    UInt32(u32),
+    UInt64(u64),
+    Int8(i8),
+    Int16(i16),
+    Int32(i32),
+    Int64(i64),
+    Int128(i128),
+    Float(SerializableOrderedF64),
+    String(Arc<str>),
+    Bool(bool),
+    Date32(i32),
+    Decimal128(i128),
+    Serialized(Arc<[u8]>),
+}
+
+impl std::fmt::Display for Value {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::UInt8(x) => write!(f, "{x}(u8)"),
+            Self::UInt16(x) => write!(f, "{x}(u16)"),
+            Self::UInt32(x) => write!(f, "{x}(u32)"),
+            Self::UInt64(x) => write!(f, "{x}(u64)"),
+            Self::Int8(x) => write!(f, "{x}(i8)"),
+            Self::Int16(x) => write!(f, "{x}(i16)"),
+            Self::Int32(x) => write!(f, "{x}(i32)"),
+            Self::Int64(x) => write!(f, "{x}(i64)"),
+            Self::Int128(x) => write!(f, "{x}(i128)"),
+            Self::Float(x) => write!(f, "{}(float)", x.0),
+            Self::String(x) => write!(f, "\"{x}\""),
+            Self::Bool(x) => write!(f, "{x}"),
+            Self::Date32(x) => write!(f, "{x}(date32)"),
+            Self::Decimal128(x) => write!(f, "{x}(decimal128)"),
+            Self::Serialized(x) => write!(f, "<len:{}>", x.len()),
+        }
+    }
+}
+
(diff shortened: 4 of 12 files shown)


optd-ng-kernel/src/cascades.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2023-2024 CMU Database Group
+//
+// Use of this source code is governed by an MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+pub mod memo;
+pub mod naive_memo;
+pub mod optimizer;
+pub mod persistent_memo;
```
connortsui20, line 9: Every file needs module-level documentation. And following the standard of most modern Rust open-source projects, this should be in `src/cascades/mod.rs`, _especially_ since it is just making its submodules public.

optd-ng-kernel/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+tracing = "0.1"
+ordered-float = "4"
+itertools = "0.13"
+serde = { version = "1.0", features = ["derive", "rc"] }
+chrono = "0.4"
+sqlx = { version = "0.8", features = [
+    "runtime-tokio",
+    "sqlite",
+] } # TODO: strip the features, move to another crate
+serde_json = { version = "1" } # TODO: move to another crate
```
connortsui20, line 24: We should strive to make our dependency versioning consistent. Ideally, we follow minimal versions (`-Zminimal-versions`), but I can understand that might be tedious. At the very least, we should choose something consistent.

optd-ng-kernel/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+
+[dev-dependencies]
+tokio = { version = "1", features = ["full"] }
```
connortsui20, line 27: Do we really need all features? `tokio` is massive and takes a while to compile.

skyzh: we can drop them in the future, this is a MVP

optd-ng-kernel/src/cascades.rs
skyzh: well, most of them were `pub(crate)` in the old codebase

optd-ng-kernel/src/cascades/optimizer.rs
```diff
@@ -0,0 +1,33 @@
...
+use std::fmt::Display;
+
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug, Default, Hash)]
+pub struct GroupId(pub(super) usize);
+
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug, Default, Hash)]
+pub struct ExprId(pub usize);
+
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug, Default, Hash)]
+pub struct PredId(pub usize);
```
connortsui20, line 15: These all need documentation, even if it is just a single line. Someone reading this might have no idea what `PredId` actually represents.

skyzh:
The next steps:

We should split the SQL migration things into a new crate, and potentially use a ORM, though I think an ORM is too heavy in our case -- I don't think we want to worry about schemas changes of optd-core for now.

On the memo table side, we can continue adding new functionalities: logical props, winner, etc.

And, with the current memo table, we can already implement 4 out of 5 cascade tasks (optimize input needs to know and update the winner)

Plus, I think we can start documenting whatever we have now + cut down the features we need on these crates. i.e., we should only use sqlx `any` or `impl Connection` interface in the core, and let the cli to decide which database backend to use.

There are also a few optimization opportunities in the memo table -- creating indexes, better merge group (can we really do lazy merging on SQL?)

optd-ng-kernel/src/cascades/optimizer.rs
```diff
@@ -0,0 +1,33 @@
...
+
+impl Display for GroupId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "!{}", self.0)
+    }
+}
```
connortsui20, line 21: Is there a reason why there is an exclamation mark here? Shouldn't this just be something like:

```rust
impl Display for GroupId {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        write!(f, "GroupId: {}", self.0)
    }
}
```

If there is a reason, then it should be documented.

optd-ng-kernel/src/nodes.rs
```diff
@@ -0,0 +1,370 @@
+// Copyright (c) 2023-2024 CMU Database Group
+//
+// Use of this source code is governed by an MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
+//! The RelNode is the basic data structure of the optimizer. It is dynamically typed and is
+//! the internal representation of the plan nodes.
```
connortsui20, line 7: This is a good start, but this could be much more specific. What else does this file contain?

optd-ng-kernel/src/tests/common.rs
```diff
@@ -0,0 +1,292 @@
...
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    cascades::optimizer::GroupId,
+    nodes::{
+        ArcPlanNode, ArcPredNode, NodeType, PersistentNodeType, PlanNode, PlanNodeOrGroup,
+        PredNode, Value,
+    },
+};
+
```
connortsui20, line 17: The logic in this file seems pretty important... is there a reason why it is in `src/tests/common` and not some normal submodule?

Also, generally we use `tests` outside of the `src` directory for integration tests.

optd-ng-kernel/src/cascades/optimizer.rs
skyzh: the old convention from optd-core that makes dump pretty...

optd-ng-kernel/src/tests/common.rs
skyzh: it's a demo repr for testing the core, it will be used in cascades/heuristics/memo unit tests

optd-ng-kernel/src/tests/common.rs
```diff
@@ -0,0 +1,292 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+impl std::fmt::Display for MemoTestPredTyp {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
```
connortsui20, line 55: Is there a reason that we need to implement `Display` for everything instead of just deriving `Debug`?

optd-ng-kernel/src/tests/common.rs
```diff
@@ -0,0 +1,292 @@
...
+// TODO: move this into nodes.rs?
+#[derive(Clone, Debug, Hash, PartialEq, Eq, Serialize, Deserialize)]
+pub struct PersistentPredNode<T: PersistentNodeType> {
+    /// A generic predicate node type
+    pub typ: T::PredType,
+    /// Child predicate nodes, always materialized
+    pub children: Vec<PersistentPredNode<T>>,
+    /// Data associated with the predicate, if any
+    pub data: Option<Value>,
+}
```
connortsui20, line 77: The documentation on this struct needs to be more specific. "A generic predicate node type" doesn't really say anything beyond what `pub typ: T::PredType` says, and it doesn't say what the purpose of this struct is for.

Edit: See https://github.com/cmu-db/optd/pull/251#issuecomment-2502532804

optd-ng-kernel/src/tests/common.rs
```diff
@@ -0,0 +1,292 @@
+// Copyright (c) 2023-2024 CMU Database Group
+//
+// Use of this source code is governed by an MIT-style license that can be found in the LICENSE file or at
+// https://opensource.org/licenses/MIT.
+
```
connortsui20, line 5: We need documentation on what is in this file. Even when reading over it, I'm not sure why the things in here should be considered "common"

optd-ng-kernel/src/tests/common.rs
```diff
@@ -0,0 +1,292 @@
...
+            Self::Project | Self::Scan | Self::Join | Self::Sort | Self::Filter
+        )
+    }
+}
+
+// TODO: move this into nodes.rs?
+#[derive(Clone, Debug, Hash, PartialEq, Eq, Serialize, Deserialize)]
+pub struct PersistentPredNode<T: PersistentNodeType> {
+    /// A generic predicate node type
+    pub typ: T::PredType,
```
connortsui20, line 72: Also, most open source projects I see that have type tags use the word "kind" instead of "typ", and I personally thing that reads better and make more sense.

optd-ng-kernel/src/tests/common.rs
```diff
@@ -0,0 +1,292 @@
...
+        matches!(
+            self,
+            Self::Project | Self::Scan | Self::Join | Self::Sort | Self::Filter
+        )
+    }
+}
+
+// TODO: move this into nodes.rs?
+#[derive(Clone, Debug, Hash, PartialEq, Eq, Serialize, Deserialize)]
+pub struct PersistentPredNode<T: PersistentNodeType> {
```
connortsui20, line 70: Are predicate nodes absolutely necessary for an minimum viable product? If they are, then I can understand (and it should be documented why it is absolutely necessary). If not, then is it possible to either remove this (temporarily, of course), or is it possible to move this logic into some other submodule? This might make more sense if there are detailed module-level docs.

optd-ng-kernel/src/tests/common.rs
```diff
@@ -0,0 +1,292 @@
...
+            typ: node.typ,
+            children: node
+                .children
+                .into_iter()
+                .map(|x| x.as_ref().clone().into())
+                .collect(),
+            data: node.data,
+        }
+    }
+}
```
connortsui20, line 105: High-level documentation on why these should be implemented would be nice.

optd-ng-kernel/src/nodes.rs
```diff
@@ -0,0 +1,370 @@
...
+
+use arrow_schema::DataType;
+use chrono::NaiveDate;
+use ordered_float::OrderedFloat;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+
+use crate::cascades::optimizer::GroupId;
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct SerializableOrderedF64(pub OrderedFloat<f64>);
```
connortsui20, line 21: What is this struct for?

optd-ng-kernel/src/nodes.rs
```diff
@@ -0,0 +1,370 @@
...
+/// keyword in rust.
+///
+/// If you want to perform conversions, use the `to_*()` functions.
+impl Value {
+    pub fn as_u8(&self) -> u8 {
+        match self {
+            Value::UInt8(i) => *i,
+            _ => panic!("Value is not an u8"),
+        }
+    }
```
connortsui20, line 97: All of this boilerplate (above and below) should be moved into a submodule.

optd-ng-kernel/src/nodes.rs
```diff
@@ -0,0 +1,370 @@
...
+                }
+                _ => panic!("{self} could not be converted into an Date32"),
+            }),
+            _ => unimplemented!("Have not implemented convert_to_type for DataType {typ}"),
+        }
+    }
+}
+
+pub trait NodeType:
+    PartialEq + Eq + Hash + Clone + 'static + Display + Debug + Send + Sync
```
connortsui20, line 218: I feel like this could just be `Copy` and most of these bounds can just go away... Either that or `Into<i16>` or something similar?

optd-ng-kernel/src/nodes.rs
```diff
@@ -0,0 +1,370 @@
...
+{
+    type PredType: PartialEq + Eq + Hash + Clone + 'static + Display + Debug + Send + Sync;
+
+    fn is_logical(&self) -> bool;
+}
+
+pub trait PersistentNodeType: NodeType {
+    fn serialize_pred(pred: &ArcPredNode<Self>) -> serde_json::Value;
+
+    fn deserialize_pred(data: serde_json::Value) -> ArcPredNode<Self>;
```
connortsui20, line 228: I feel like there should be different methods on these. Going back to a previous comment, do we absolutely need predicates for the MVP? And then shouldn't there be methods to serialize the entire node instead of just the type tag and the predicate?

optd-ng-kernel/src/nodes.rs
```diff
@@ -0,0 +1,370 @@
...
+    fn serialize_plan_tag(tag: Self) -> serde_json::Value;
+
+    fn deserialize_plan_tag(data: serde_json::Value) -> Self;
+}
+
+/// A pointer to a plan node
+pub type ArcPlanNode<T> = Arc<PlanNode<T>>;
+
+/// A pointer to a predicate node
+pub type ArcPredNode<T> = Arc<PredNode<T>>;
```
connortsui20, line 239: This makes things harder to read, and also harder to traverse with an LSP.

optd-ng-kernel/src/nodes.rs
```diff
@@ -0,0 +1,370 @@
...
+
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+pub enum PlanNodeOrGroup<T: NodeType> {
+    PlanNode(ArcPlanNode<T>),
+    Group(GroupId),
+}
```
connortsui20, line 245: Missing documentation.

optd-ng-kernel/src/nodes.rs
```diff
@@ -0,0 +1,370 @@
...
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+pub struct PlanNode<T: NodeType> {
+    /// A generic plan node type
+    pub typ: T,
+    /// Child plan nodes, which may be materialized or placeholder group IDs
+    /// based on how this node was initialized
+    pub children: Vec<PlanNodeOrGroup<T>>,
+    /// Predicate nodes, which are always materialized
+    pub predicates: Vec<ArcPredNode<T>>,
+}
```
connortsui20, line 292: This needs _**highly detailed**_ documentation. Considering the importance of this struct, someone reading this codebase should only have to look at this single struct and be able to understand _everything_ about optd's operator / plan representation.

skyzh:
I don't know how detailed you want -- could you please help document `PlanNode` to set a common expectations for us?

connortsui20 (changes requested):
Ok I'm going to stop myself because I feel I am getting too repetitive and it's not really conducive to good progress, and I feel myself getting too opinionated.

I think that this is a great start, but considering we want to build a kernel of a foundation, we have to hold the code quality to a very high standard. This means that _everything_ needs to be documented.

I think that documentation will help a lot with readability, and it will probably also help us identify things that we can probably do to minimize the MVP. Do we need separate predicate representation for an MVP? Should we even have generic types at all? Should plan nodes be able to store their own fields to facilitate easy serialization and deserialization?

I think the biggest thing we need to worry about is what this would look like when backed by a database via an ORM. I think it makes a lot more sense to just allow plan nodes to have whatever fields they want as it means getting all of the information about a single expression is a single look + deserialization. By storing children and predicates as vectors, we necessitate _multiple_ joins just to get information about a single expression.

If it is at all possible to minimize this even further, acknowledging that we will be losing functionality, that is what we should strive for with this. Aim for the smallest thing possible.

connortsui20:
> I don't know how detailed you want -- could you please help document `PlanNode` to set a common expectations for us?

Every single field should have an explanation on why we need it to be there and how code should interact with it. For example, I had no idea that `typ: T` is a generic type tag that other developers are allowed to choose. I thought for the longest time that the generic type _WAS_ the tag. That is a subtle difference that is very not obvious, because when you see a type like `Foo<T>`, you expect it to have some sort of "wrapping" behavior around `T` (think `Vec<T>` or `Option<T>`), but that is not what is happening here.

Why do the children and predicates need to be vectors? What does "materialized" actually mean? I now know what it means, but if someone comes along and looks at it, they would have no idea.

These are some of the things that should be written down. Or if it is already written down, there should a link to where it is.

optd-ng-kernel/Cargo.toml
connortsui20: Fair enough, but there should be a TODO comment saying that it should be changed.

optd-ng-kernel/src/cascades/optimizer.rs
connortsui20: That should be documented! Even if this was in the old repo, it's not obvious that this is the purpose.

PR #252, opened 2024-12-05
Filter/Agg transpose bugfix

jurplel:
The check for whether or not a filter expression could be pushed beyond an agg node was incorrect. 

It was checking if the column was in the group by columns (checking the equality of the numbers), when it should be checking based on indices if we are only referring to columns that are *emitted* from the agg node as group by columns.

For example, if we see:
```
Filter #1 > 100
    Agg { groups: [#1], agg: Sum() }
```

We should *not* push down because `#1` refers to the sum column. In the current main branch, it is pushed down because it sees that `#1` equals a column in the `groups` field. It should be checking that every column is `< groups.len()` instead.


skyzh (approved):

optd-sqllogictest/slt/unnest-dup.slt
```diff
@@ -0,0 +1,9 @@
+include _basic_tables.slt.part
+
+query
+select * from t1 where (select sum(v4) from t2 where v3 = v1) > 100;
+----
+2 200
+2 250
+3 300
+3 300
```
yliang412: nit: newline

optd-datafusion-repr/src/rules/filter_pushdown.rs
```diff
@@ -346,7 +346,7 @@ fn apply_filter_agg_transpose(
         let mut group_by_cols_only = true;
         for child in children {
             if let Some(col_ref) = ColumnRefPred::from_pred_node(child.clone()) {
-                if !group_cols.contains(&col_ref.index()) {
```
yliang412, line 349: Could we add some comments on `col_ref.index() >= group_cols.len()` and how the column indices dependency works? It is not a super obvious thing.

yliang412: This is implicit b/c datafusion Agg schema is group columns followed by the agg columns.

PR #254, opened 2024-12-06
Fix for TPC-H Q11

jurplel:
Test failed with `test panicked: Column index 2 out of bounds 2 + 0`

The 0 schema len was due to this part of the plan:
```
            └── LogicalProjection
                ├── exprs:Cast
                │   ├── cast_to: Decimal128(38, 15)
                │   ├── child:Mul
                │   │   ├── Cast { cast_to: Float64, child: #0 }
                │   │   └── 0.0001(float)
```

Turns out, `BinOp` was being considered to have an empty schema length. 

Let me know if this fix does not take into account some case.


skyzh:
binop should have schema length = 1 because it’s a single column output?

skyzh:
All expressions should have exactly one Field::placeholder

skyzh:
I feel a lot of things are very wrong here... The schema is expected to be the output type of a predicate. Therefore, the default branch should always return a schema with one field, so do most of the other predicates. Pushed a fix.

jurplel:
Thanks for the fix!

PR #255, opened 2024-12-06
Fixes for TPC-H Q17

jurplel:
Two bugs:
- `dep_initial_distinct` did not work if the correlated columns were not `[#0, #1, #2]` and so on.
- `dep_join_past_agg` was pushing the actual correlated column number instead of the *index* of the correlated column number. This works in the same cases the above bug did (`[#0, #1, #2]` correlated columns), and so it broke for the same reason afaict.


skyzh (approved):
can you please also add a minimum reproducible test case for this?

PR #257, opened 2024-12-07
Subquery Unnesting Agg NULL case workarounds

jurplel:
- Add an outer join with the deduplicated "left side", and a corresponding projection node, to pass along NULL values as expected.
- Add a specific workaround in the projection node to case NULL -> 0 if we have a COUNT(*).


optd-datafusion-repr/src/rules/subquery/depjoin_pushdown.rs
```diff
@@ -345,11 +343,87 @@ fn apply_dep_join_past_agg(
...
+
+    // We have to maintain the same schema above outer join as w/o it, but we
+    // also need to use the groups from the deduplicated left side, and the
+    // exprs from the new agg node. If we use everything from the new agg,
+    // we don't maintain nulls as desired.
+    let outer_join_proj = LogicalProjection::new(
+        new_outer_join.into_plan_node(),
+        ListPred::new(
+            (0..left_schema_size)
+                .chain(left_schema_size + left_schema_size..left_schema_size + new_agg_schema_size)
```
jurplel, line 397: This is one thing I am not sure of the correctness of.

The meaning is to take:
- Everything from the left side
- Everything from the right side that *is not in the left side*.

I found this to be correct in all the tests that I wrote.

skyzh: > Everything from the right side that is not in the left side.

Maybe we need to figure out the exact meaning of this -- do you want to know if two columns refer to the same table-column?

skyzh (approved):
...anyways, let's get thing through as long as they fix things, even if it's not 100% fix, and visit them back later

optd-datafusion-repr/src/rules/subquery/depjoin_pushdown.rs
jurplel: Okay, so after thinking about it a bit longer, my description is correct, but it is a bit easier to think about than that:

The right side is going to be the subquery on the right, and the deduplicated part on the left.

The left side is going to be a copy of the deduplicated part that is the left child of the right child.

So, our projection will be the deduplicated copy (that will be outer joined to resolve the null problems) and the non-deduplicated part from the right side.

PR #260, opened 2024-12-07
feat(df-repr/bridge): upgrade datafusion to 43.0.0

skyzh:
Despite the upgrade,

* New `create_df_context` to be used across all crates to create a datafusion context with optd. We had too much duplicate code before to set up the context.
* The main refactor is about the aggregation expressions. Datafusion has a new way of doing that.
* Datafusion removed cross join. We didn't. We can eventually remove it but now it's blocked on two-stage cascades: if we simply treat cross join the same as inner join, we would time out.
* Several other refactors to adapt to datafusion (i.e., limit node now takes i64, empty relation / placeholder row executor)
* Keep as much as the original datafusion cli crate as possible. We now only patch main.rs and exec.rs.
* There's one more breaking change that we might encounter later when doing sort physical properties. Now datafusion logical plan will remove duplicate sorts if there are no limits present. I feel this is a bad move b/c it's not a direct mapping from the original SQL statement...


jurplel:
Waiting to review until after sqllogictest failures are fixed

jurplel (approved):

PR #261, opened 2024-12-09
Extra decimal support for TPC-H Q18

jurplel:
Regrettably the query is still too slow...

Would like a sanity check on this in any case.

Won't merge until unnest-exists is merged


optd-core/src/nodes.rs
```diff
@@ -211,7 +211,16 @@ impl Value {
                 }
                 _ => panic!("{self} could not be converted into an Date32"),
             }),
-            _ => unimplemented!("Have not implemented convert_to_type for DataType {typ}"),
+            DataType::Decimal128(_, _) => Value::Decimal128(match self {
+                // TODO: Should we be ignoring the scale and precision here?
```
skyzh, line 215: I think Postgres has a set of rules to determine decimal in this case but let's ignore it for now...

skyzh (approved):

PR #259, opened 2024-12-07
Subquery Unnesting: Exists + In Support

jurplel:
- Support uncorrelated/correlated IN (ANY) and EXISTS clauses (This should effectively make subquery unnesting feature complete!)
- TPC-H Q4, Q16, Q20, and Q22 working
    - The remaining queries mostly seem to have plans that are too bad (Q18 has some other issue, [fix in the pipeline](https://github.com/cmu-db/optd/pull/261) but it still seems like the plan is too slow).

PR #264, opened 2024-12-10
sqlplannertest: remove duplicated yaml that were disabled

yliang412:
https://github.com/cmu-db/optd/pull/259 added new yml test cases instead of renaming the `.disabled` ones, removing them now.


jurplel (approved):
My bad—thank you!

PR #265, opened 2024-12-14
feat(core/df-repr): better upper bound pruning to reduce plan space

skyzh:
* Remove cross join from the repr
* Implement upper bound pruning correctly
* Two-stage cascades, also use the new task graph instead of refactoring the current one


yliang412 (commented):
Reviewed:
- cross join to inner join changes
- main upper bound pruning logic in `optimize_inputs_inner`

I need another look at `task2.rs` and potentially doc more aggressively like https://github.com/cmu-db/optd/pull/234. I actually like this version of the task graph a lot (no state machine logic).

Is there any other part of the logic I need pay special attention to?

skyzh:
I think part of this PR is tuning to ensure the plan is what we want (i.e., filter selectivity and behavior on reaching the budget). The tasks2.rs logic is also something that we should fully understand as we are going to bring more logic into it (phys prop, etc.)

yliang412 (approved):
Needs to doc the task graph a little bit more but LGTM

PR #266, opened 2024-12-17
test: reproduce hash join planning issue

skyzh:


skyzh:
included in https://github.com/cmu-db/optd/pull/265

PR #274, opened 2024-12-22
feat: support left-outer and left-mark hash join impl rules

yliang412:
## Problem

We should be able to convert left-outer join and left-mark logical equi join to hash join.

## Summary of changes

- Add implementation rules to handle these cases.
- Add join-split-filter rules that extract predicates in the join condition into those that can be pushed down as filters.

_misc_
- refactor `simplify_log_expr` to stop using `unreachable!`


yliang412:
tpch Q13 needs a rule to push split a filter from the join node, and then the join could be turned into a left-outer hash join. Working on this now.


<pre>
...
  ├── cond:And
  │   ├── Eq
  │   │   ├── #0
  │   │   └── #9
  │   <b>└── Like { expr: #16, pattern: "%special%requests%", negated: true, case_insensitive: false }</b>
</pre>

optd-sqlplannertest/tests/tpch/q13.planner.sql
```diff
@@ -61,14 +61,9 @@ PhysicalSort
...
-            ├── join_type: LeftOuter
-            ├── cond:And
-            │   ├── Eq
-            │   │   ├── #0
-            │   │   └── #9
-            │   └── Like { expr: #16, pattern: "%special%requests%", negated: true, case_insensitive: false }
+        └── PhysicalHashJoin { join_type: LeftOuter, left_keys: [ #0 ], right_keys: [ #1 ] }
             ├── PhysicalScan { table: customer }
-            └── PhysicalScan { table: orders }
+            └── PhysicalFilter { cond: Like { expr: #8, pattern: "%special%requests%", negated: true, case_insensitive: false } }
```
yliang412, line 66: pushing down the filter and turn into hash join.

optd-sqlplannertest/tests/tpch/q16.planner.sql
```diff
@@ -91,11 +91,7 @@ PhysicalSort
         │   ├── InList { expr: Cast { cast_to: Int64, child: #10 }, list: [ 49(i64), 14(i64), 23(i64), 45(i64), 19(i64), 3(i64), 36(i64), 9(i64) ], negated: false }
         │   └── Not
         │       └── [ #14 ]
-        └── PhysicalNestedLoopJoin
-            ├── join_type: LeftMark
-            ├── cond:Eq
-            │   ├── #1
-            │   └── #14
+        └── PhysicalHashJoin { join_type: LeftMark, left_keys: [ #1 ], right_keys: [ #0 ] }
```
yliang412, line 94: picked up by the new hash-join-left-mark rule

optd-sqlplannertest/tests/tpch/q17.planner.sql
```diff
@@ -78,52 +78,48 @@ PhysicalProjection
...
-                                │   └── [ #5 ]
-                                ├── groups: [ #0 ]
-                                └── PhysicalHashJoin { join_type: Inner, left_keys: [ #0 ], right_keys: [ #1 ] }
-                                    ├── PhysicalAgg { aggrs: [], groups: [ #16 ] }
-                                    │   └── PhysicalNestedLoopJoin { join_type: Inner, cond: true }
-                                    │       ├── PhysicalScan { table: lineitem }
-                                    │       └── PhysicalScan { table: part }
-                                    └── PhysicalScan { table: lineitem }
+                └── PhysicalProjection { exprs: [ #0, #2 ] }
+                    └── PhysicalHashJoin { join_type: LeftOuter, left_keys: [ #0 ], right_keys: [ #0 ] }
```
yliang412, line 109: left-outer hash join

optd-sqlplannertest/tests/tpch/q2.planner.sql
```diff
@@ -262,12 +262,7 @@ PhysicalLimit { skip: 0(i64), fetch: 100(i64) }
             │       │   └── "AFRICA"
             │       └── PhysicalScan { table: region }
             └── PhysicalProjection { exprs: [ #0, #2 ] }
-                └── PhysicalNestedLoopJoin
-                    ├── join_type: LeftOuter
-                    ├── cond:And
-                    │   └── Eq
-                    │       ├── #0
-                    │       └── #1
+                └── PhysicalHashJoin { join_type: LeftOuter, left_keys: [ #0 ], right_keys: [ #0 ] }
```
yliang412, line 265: left-outer hash join

optd-sqlplannertest/tests/tpch/q20.planner.sql
```diff
@@ -106,84 +106,71 @@ PhysicalSort
         │   └── Eq
         │       ├── #8
         │       └── "IRAQ"
-        └── PhysicalNestedLoopJoin
-            ├── join_type: LeftMark
-            ├── cond:Eq
-            │   ├── #0
-            │   └── #11
+        └── PhysicalHashJoin { join_type: LeftMark, left_keys: [ #0 ], right_keys: [ #0 ] }
```
yliang412, line 109: left-mark hash join

optd-sqlplannertest/tests/tpch/q20.planner.sql
```diff
@@ -106,84 +106,71 @@ PhysicalSort
...
+                    │   ├── Gt
+                    │   │   ├── Cast { cast_to: Float64, child: #2 }
+                    │   │   └── #8
+                    │   ├── Eq
+                    │   │   ├── #0
+                    │   │   └── #6
+                    │   └── Eq
+                    │       ├── #1
+                    │       └── #7
+                    ├── PhysicalFilter { cond: #5 }
```
yliang412, line 126: This one is a little confusing. investigating ...

optd-datafusion-repr/src/rules/filter_pushdown.rs
```diff
@@ -160,6 +160,87 @@ fn apply_filter_merge(
...
 }
 
+// Rule to split predicates in a join condition into those that can be pushed down as filters.
+define_rule!(
+    JoinInnerSplitFilterRule,
+    apply_join_split_filter,
+    (Join(JoinType::Inner), child_a, child_b)
+);
+
+define_rule!(
```
skyzh, line 170: I don't think this rule is correct. You cannot move the outer join condition into a filter in some cases.

Consider `select * from a left join b on a.x = b.y and b.z = 1`. The result is different from `select * from a left join b on a.x = b.y where b.z = 1`. Assume left table is x=1, right table is y=1,z=2, the correct result is 1, NULL, NULL, versus the rule will produce zero rows.

skyzh: ahh, I realized that this is a filter pushdown, then it might be correct; I will do a review later :)

